PR #30, opened 2022-07-28
Add index features, format references

GloriaWYY:


trevorcampbell:
The deploy failed, but it looks unrelated to the changes. The PR looks good to me. Merging. We can deal with the deploy separately.

PR #44, opened 2022-12-16
Update metadata and add chapter numbering

joelostblom:
You can use the link below (https://deploy-preview-44--mellifluous-pie-f3e391.netlify.app/) to review any changes introduced in this PR. You can see that the change in the course title and copyright year is reflected there. I like this because it makes it quicker to review.

For now I turned off auto execution of notebooks because there was a dependency issue on Netlify, but if we like this review approach I can turn it back on and figure out the dependencies.


trevorcampbell:
Thanks for investigating! I'm kind of lukewarm on this. It basically doesn't provide any advantage over just running `built_html.sh` yourself locally aside from the fact that the PR is prebuilt, so you don't have to checkout the branch and wait for the build when reviewing the PR. I'm not totally against it either though, it is a small convenience.

Can you provide more specific detail on how you set this up to trigger on PRs / what triggers it?

> For now I turned off auto execution of notebooks because there was a dependency issue on Netlify,

not sure what "autoexecution of notebooks" is in this context

trevorcampbell:
(Assuming this triggers on a PR to `main`, we should switch the target branch to `dev`, since all development work will happen there from now on, to separate development and production versions)

joelostblom:
I agree it is mostly a small convenience. I'm generally find that I require less activation energy when I can just click a link to  see the built side versus building it myself, especially if I go back and fourth with someone. I'm OK either way, but would lean towards having it.  I guess it doesn't hurt to have it set up, we can still check out locally if we prefer that workflow.

> not sure what "autoexecution of notebooks" is in this context

By default JupyterBook automatically executes notebooks during build-time. This behavior can either be to always execute all notebooks, execute only the ones with changes (worked just ok when I tried it a while back), or don't execute at all and use the outputs from the last saved state. To have quicker build times I usually go for not executing at all when I built my lecture sites for the other courses, and make sure that I click "restart and run all" in my lecture notebooks if I make substantial changes. It is of course more reproducible to execute notebook on build, but we could also have a separate test for this so I am OK either way here too (as long as we have some check that the notebook runs from top to bottom).

trevorcampbell:
> It is of course more reproducible to execute notebook on build, but we could also have a separate test for this so I am OK either way here too (as long as we have some check that the notebook runs from top to bottom).

Yes, in my experience, you cannot rely on others to be careful about doing that sort of thing themselves. You have to use build tools that handle it automatically. So if you could get that working that would be great.

Also, it seems this tool on my other PRs is failing. No idea why, because the deploy log isn't loading. Any idea?

(the lack of logs to diagnose failures, plus all the email spam about failures from this tool are starting to push me towards "get rid of it"...)


Screenshot from 2022-12-19 12-04-50

joelostblom:
Yes the reason is the depency I mention for executing the notebooks. Until
I have time to look into that, it will only work when we turn the execution
off as I have done in this PR. But that doesn't work now that we use md as
the source format so I will try to look into fixing the depency but might
not be until I get back to Sweden, just landed in Calgary. I can turn it
off in the meanwhile.

joelostblom:
I think I have fixed the build issues so this should work with all PRs now and we can merge it if you're OK with it @trevorcampbell .

I had to update our `requirements.txt` since netlify automatically picks up on the packages there. Due to this issue I had to install jupyterbook from github: https://github.com/executablebooks/jupyter-book/issues/1898

I also updated the ToC so that it looks nicer, partly to have something to test with.

These are the only setting on Netlify now:

image

requirements.txt
```diff
@@ -1,3 +1,4 @@
-jupyter-book
-matplotlib
-numpy
+git+https://github.com/executablebooks/jupyter-book
+altair_saver
```
trevorcampbell: Can we just delete this file now? Our build environment is entirely in docker now

source/_config.yml
```diff
@@ -3,9 +3,9 @@
 #######################################################################################
 # Book settings
 
-title: DSCΙ 100
+title: DSCΙ 100 Py
```
trevorcampbell: Title is

Data Science: A First Introduction (Python Edition)

source/_config.yml
```diff
@@ -3,9 +3,9 @@
...
 author: UBC
-copyright: "2021" # Copyright year to be placed in the footer
+copyright: "Since 2022" # Copyright year to be placed in the footer
```
trevorcampbell: remove "since"

source/_config.yml
```diff
@@ -75,9 +75,9 @@ launch_buttons:
 
 
 repository:
-  url: https://github.com/phaustin/eosc211_students # The URL to your book's repository
+  url: https://github.com/UBC-DSCI/introduction-to-datascience-python # The URL to your book's repository
   path_to_book: "" # A path to your book's folder, relative to the repository root.
-  branch: e211_live_main # Which branch of the repository should be used when creating links
+  branch: main # Which branch of the repository should be used when creating links
```
trevorcampbell: Is it `main` or `production` or `gh-pages` we should use here?

trevorcampbell (changes requested):
@joelostblom minor changes needed

requirements.txt
joelostblom: Yup I think so

source/_config.yml
joelostblom:
```suggestion
title: Data Science: A First Introduction (Python Edition)
```

source/_config.yml
```diff
@@ -3,9 +3,9 @@
 #######################################################################################
 # Book settings
 
-title: DSCΙ 100
+title: Data Science: A First Introduction (Python Edition)
 author: UBC
-copyright: "2021" # Copyright year to be placed in the footer
+copyright: "Since 2022" # Copyright year to be placed in the footer
```
joelostblom:
```suggestion
copyright: "2022" # Copyright year to be placed in the footer
```

source/_config.yml
joelostblom: I went with production for now, I am not 100% sure

source/_config.yml
joelostblom: Removed, although I think we will have to update it every year then, won't we? I usually see something like 2022-23 for example.

joelostblom:
@trevorcampbell  Ready for a review again. I can't find a way to remove the binder button, but that can be another PR

source/_config.yml
trevorcampbell: nope -- specifying a copyright date just refers to when the work was first created. it does not need to be updated

(honestly specifying a copyright date is pointless anyway -- copyright law in basically any jurisdiction applies from the first provable date of creation anyway, regardless of notices posted)

source/_config.yml
trevorcampbell: my guess is that if we are creating links files in `source/`, the `production` branch makes the most sense. If we are creating links to files in `source/_build`, then `gh-pages` makes sense. I don't think `main` would be used.

trevorcampbell (approved):
LGTM now! Will merge.

PR #48, opened 2022-12-17
Production polish chapter 1

### LINKED ISSUE #13: intro
The deployed notebook can be viewed at : https://nd265.github.io/introduction-to-datascience-python/intro.html

### LINKED ISSUE #34: Explain filtering of rows and selection of columns in a more Pandas-centric way
The introduction of `[]` for row filtering and `.loc[]` for column filtering seems like a translation of `filter` and `select` from the tidyverse, but in Pandas `[]` can be used for either columns (by names) or rows (by slices or boolean masks), whereas `.loc[]` is used when both rows and columns are to be selected at the same time to avoid ambiguous chained operations.

### LINKED ISSUE #37: Suggestions for Ch 1 - Pandas intro

- [x] "Python and THE pandas" -> "Python and pandas"
- [x] Elaborate on pd import alias
- [x] Fig 2 is the R-version and the font is too small
- [x] Remove print statements under "Naming things"
- [x] Variable names can't use `.`
- [x] Switching to `-` instead of `+` is a more natural example for non-working var names.
- [x] Data frames (>60 rows) are not printed just as the first few rows in Python, it also include the last few rows and a `...` that could use an explanation. And the number of rows and columns are printed at the bottom, not top.
- [x] Mention that single and double quotes are exchangeable to avoid confusion
- [x] The corresponding figures to 1.3 and up, in the R book are all missing in python.
- [x] Should we explain `.query` or `[]`? The former has some advantages in clearer chaining syntax, but is not as well documented.
- [x] "greatest to least" -> "largest to smallest value"
- [x] Introduce the contrast of df.method versus pd.function. Dataframes have the most important functions built-in which is great for finding help and what you're looking for.
- [x] "introduce tidy data in Chapter Cleaning" -> "introduce tidy data in THE Chapter Cleaning"
- [x] "A bar plot is a chart where the heights of the bars" -> "A bar plot is a chart where the LENGTHS of the bars".
- [x] R - Isn't stat == identify the default for bar? Or is that column? And does it even matter with one value?
- [x] The Python explanation for how expressions can continue on another line is wrong. It is not the `.`, it is the open paren (which is preferred over "\")
- [x] Better explanation of alt.X and alt.Y, take from 531 lecture notes.
- [x] I think the organization of the bars could benefit from having the largest bar closest to the axis.
- [x] Visualization syntax in "putting it all together" is off
- [x] Discuss the most common ways of getting help in JupyterLab, shift+tab and the contextual explorer. `?` also works in place of `help` in ipython, I don't think printing the docstring is worth teaching.
- [x] Exercises mention "R and the tidyverse"
- [ ] https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/35
- [ ] https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/34
- [ ] https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/29

trevorcampbell:
~~Merge #46 before this one.~~ #46 has been merged, so this is safe now.

- closes #13 
- closes #37
- closes #34


source/intro.md
```diff
@@ -453,82 +456,114 @@ to only those languages in the "Aboriginal languages" category.
...
+> a data frame (as in `.loc[...]`) *and* to denote a list (as in `["language", "mother tongue"]`).
+> You could, for example, give the list its own name before selecting the columns in the data frame.
+```{code-cell} ipython3
+column_names = ["language", "mother_tongue"]
+selected_lang = aboriginal_lang.loc[:, column_names]
+selected_lang
 ```
 
-We have used `df[]` and `loc[]` properties of dataframe to obtain a table with only the Aboriginal
+### Using `sort_values` to order and `.iloc[]` to select rows by index number
```
joelostblom: As a cosmetic comment, I think we should stick to either using `.` for all attributes (so also `.sort_values`) or for none of them. I'm OK with either but leaning towards the latter as the `.` seems redundant and is not adding anything (that I can think off). I'm guessing it is also less work, since in the R versions there are no `.` and most chapters are probably following that structure right now.

trevorcampbell: Agree -- will remove the leading `.` on my next pass through this one

trevorcampbell: fixed

source/intro.md
```diff
@@ -417,131 +418,195 @@ Wait a minute, nothing happened this time! Where's our data?
...
+
+We can use the `[]` operation to extract the `category` column from the `can_lang` data frame.
+```{code-cell} ipython3
+can_lang["category"]
 ```
+Notice how the single column here looks a little different from the data frame we printed out earlier!
+For example, there is no column title listed at the top any more---it has moved to the bottom (where it says `Name: category`).
+In fact, by running the code `can_lang["category"]`, we have pulled a single column out of the data
+frame as a *series* object. You will learn more about series later in the {ref}`wrangling` chapter,
+but for now, you can just think of this as a column that isn't inside a data frame any more.
```
joelostblom: Part of me feels like an introductory course like dsci 100 could maybe be thought without mentioning series and just always select with `[['col']]` so that a dataframe is returned. But that's not very standard and I am sure they will stumble onto the concepts of series online and be confused so I guess it is better to teach it. My biggest issues with pandas for novices are the concept of a series and the index, I wish these were opt-in mechanisms for power users instead of the default behavior.

source/intro.md
```diff
@@ -417,131 +418,195 @@ Wait a minute, nothing happened this time! Where's our data?
...
+here. The `loc[]` operation allows you to obtain a subset of *both* the rows
+and columns of a data frame at once. In order to tell the `loc[]` operation
+which subset of columns we want to obtain, we can just list their names.
+In order to tell the `loc[]` operation which subset of rows we want to obtain, 
+we can examine the values of a single column of the data frame using the `[]` operation. 
+In our example in particular, we will use 
+`[]` to look at the `category` column in `can_lang` to extract only the 
+Aboriginal languages in the data set.
+
+> **Note:** The `[]` operation is much more general than what we describe in this chapter.
```
joelostblom:
```suggestion
> **Note:** The `[]` operation is much more powerful than what we describe in this chapter.
```

source/intro.md
```diff
@@ -417,131 +418,195 @@ Wait a minute, nothing happened this time! Where's our data?
...
+and then the second step is to keep only the `language` and `mother_tongue` columns.
+The `[]` and `loc[]` operations on the `pandas` data frame will help us
+here. The `loc[]` operation allows you to obtain a subset of *both* the rows
+and columns of a data frame at once. In order to tell the `loc[]` operation
+which subset of columns we want to obtain, we can just list their names.
+In order to tell the `loc[]` operation which subset of rows we want to obtain, 
+we can examine the values of a single column of the data frame using the `[]` operation. 
+In our example in particular, we will use 
+`[]` to look at the `category` column in `can_lang` to extract only the 
+Aboriginal languages in the data set.
```
joelostblom: I think this is a little complex to understand. For example, I had to read this sentence multiple times and I think it makes it sound like `[]` has some special meaning inside `loc`: "In order to tell the `loc[]` operation which subset of rows we want to obtain, we can examine the values of a single column of the data frame using the `[]` operation. "

I suggest changing this section slightly and showing one thing at a time more slowly. I also think we can still stick to the "filter rows" and "select columns" terminology. So this section could read:

> The first step is to select the the `language` and `mother_tongue` columns from our data. Then we want to filter the data to extract only those rows that correspond to Aboriginal languages. We will first do these two steps separately and then combine them into single command. 

I have switched the order because I think it is easier to do column selection so it makes sense to see that first. Then we could follow by explaining the following:

1. Use `[]` to select the two column we want
2. Use `[]` to filter the rows of the data frame to keep only the aboriginal languages
3. How to combine the two things we just learned in a single command with `loc`.

I think this makes for a natural mental mapping of `[]` for either row filtering or column selection one at a time and `loc` for combing both, while making it clear that is the only difference they need to remember. What do you think?

source/intro.md
```diff
@@ -417,131 +418,195 @@ Wait a minute, nothing happened this time! Where's our data?
...
+
+> **Note:** The `[]` operation is much more general than what we describe in this chapter.
+> You can actually use it to select multiple columns from a data frame, or to obtain a subset of rows.
+> You will learn more about `[]` (and other indexing operations on data frames) later on
+> in the {ref}`wrangling` chapter.
+
+### Using `[]` to examine a single column
```
joelostblom: The three headings I am suggesting could be: "Using `[]` to select columns", "Using `[]` to filter rows", "Using `loc` to combine columns selection and row filtering"

source/intro.md
```diff
@@ -417,131 +418,195 @@ Wait a minute, nothing happened this time! Where's our data?
...
+a specification of which rows to select, and a specification of which columns to select.
+Let's start by just extracting the rows from `can_lang` corresponding to Aboriginal languages.
 
-+++
+**(FIGURE 1.3 FROM R BOOK MISSING HERE; BUT NOTE THAT THE STRUCTURE HERE HAS CHANGED SINCE [] and LOC[] DO NOT CORRESPOND
+TO FILTER AND SELECT. PLEASE READ THIS CHAPTER CAREFULLY BEFORE MAKING FIGS)**
 
 ```{code-cell} ipython3
-selected_lang = aboriginal_lang.loc[:, ['language', 'mother_tongue']]
+aboriginal_lang = can_lang.loc[can_lang["category"] == "Aboriginal languages", :]
```
joelostblom: I would suspect it is more intuitive if we do the selection of ["language", "mother_tongue"] first and don't introduce special syntax such as `:`. If we teach row filtering and column selection with `[]` as suggested above, then we never need to introduce `:`, since it we want all row or all columns then we use `[]` and it is only if we want to both filter columns and select rows that we use `loc` which means we never need `:`.

I think breaking it down to show both columns selection and row filtering with `[]` first will help with this so that there is not increased cognitive load from both learning new things and combining them at the same time; instead when they use `loc` they just combine two things they have already done with `[]` before.

source/intro.md
```diff
@@ -619,8 +685,8 @@ to fields in the dataset, accessed by field name(column names)
 barplot_mother_tongue = (
     alt.Chart(ten_lang)
     .mark_bar().encode(
-        x='language',
-        y='mother_tongue'
+        x="language",
+        y="mother_tongue"
     ))
```
joelostblom, line 707: My personal preference is that we use single quotes by default since they are equivalent and it is less typing but it is of course not a major point.

source/intro.md
```diff
@@ -651,13 +717,18 @@ Bar plot of the ten Aboriginal languages most often reported by Canadian residen
...
+> single Python expression span multiple lines. Above is one such case: here, Python sees that we put a left
+> parenthesis symbol `(` on the first line right after the assignment symbol `=`, and knows that our
+> expression cannot end until we close it with an appropriate corresponding right parenthesis symbol `)`. 
+> So Python keeps reading the next line to figure out
+> what the rest of the expression is.  We could, of course,
+> put all of the code on one line of code, but splitting it across
 > multiple lines helps a lot with code readability.
+> Another case where Python lets you write an expression on multiple lines is when you are using a function
+> with many arguments; you can split each argument onto its own line (terminated with a comma `,`). The trailing
+> comma tells Python to keep reading on the next line.
```
joelostblom:
```suggestion
```

This is the same as above, it is the open parenthesis in the function and not the trailing comma that tells python to keep reading.

source/intro.md
```diff
@@ -830,15 +902,20 @@ these steps in much more detail!
 can_lang = pd.read_csv("data/can_lang.csv")
 
 # obtain the 10 most common Aboriginal languages
-aboriginal_lang = can_lang[can_lang['category'] == 'Aboriginal languages']
-arranged_lang = selected_lang.sort_values(['mother_tongue'], ascending=False)
-ten_lang = arranged_lang[:10]
+ten_lang = (
+    can_lang.loc[can_lang["category"] == "Aboriginal languages", :]
```
joelostblom:
```suggestion
    can_lang[can_lang["category"] == "Aboriginal languages"]
```

I think this is a good example for when the syntax looks easier to understand when we don't use `:`

source/intro.md
```diff
@@ -871,9 +948,9 @@ Bar plot of the ten Aboriginal languages most often reported by Canadian residen
 There are many Python functions in the `pandas` package (and beyond!), and 
 nobody can be expected to remember what every one of them does
 or all of the arguments we have to give them. Fortunately, Python provides 
-the `help` function and `__doc__` attrribute, which 
+the `help` function, which 
 provides an easy way to pull up the documentation for 
-most functions quickly. To use the `help` function to access documentation, you 
+most functions quickly. To use the `help` function to access the documentation, you 
 just put the name of the function you are curious about as an argument inside the `help` function.
 For example, if you had forgotten what the `pd.read_csv` function
```
joelostblom, line 966: I do think we should mention shift+tab here or the contextual help which are the two most effective ways to get help in JuptyerLab.

joelostblom (changes requested):
Some comments! I prioritized this so that I wouldn't hold you up. Overall my biggest comments are around simplifying the `[]` / `loc[]` situation further.

source/intro.md
trevorcampbell: I'm happy to go with either, as long as we keep it consistent throughout. 

But for now, I think we are likely to introduce many inconsistencies anyway for this first pass (we're aiming for "good enough to teach from" right now). So let's not be too militant about it for now and come back and do a thorough pass through later.

I'll open an issue.

source/intro.md
trevorcampbell: ah good point, will fix

source/intro.md
trevorcampbell: I'm not against it, but I've never got either shift+tab or ctrl+I (or just right click+contextual help) to work reliably. Both most often just do nothing when I try to activate them....

Are you able to get it to work in our course container? Maybe we missed installing something in our docker container?

The nice thing about `help()` is that it works regardless of whether you're using Jupyter (Lab)

joelostblom: Yup it works fine in the course container and I was showing it to my students this year. One thing that sometimes make it seem like it is not working is that for R you have to have the cursor positioned on the function name just before the open parenthesis:

image

image

Ctrl + I / contextual help also works fine for me for both languages, I think it can be good to show this too. I tend to recommend novices to always have it open to learn as they start out. I agree it is still nice to mention `help` since it works everyone, but I don't think it is as convenient to use in the notebook.

source/intro.md
trevorcampbell: After thinking about it yesterday, I think I am on board with your idea. I've got two issues with it, one sort of unimportant, and one more important.

Unimportant: I find it aesthetically unappealing that `[ ]` is used for selecting rows, or selecting columns, and then you need a totally separate thing `.loc[ ]` to select both rows and columns. I am happy as long as we resolve this whole mess with `[ ]`, `loc[ ]`, `iloc[ ]`, series, and indices in Chapter 3.

More important: How should we best explain the following
```
can_lang[can_lang["category"] == "Aboriginal languages"]
```
*without* mentioning series?

I can imagine the appearance of `can_lang` twice in that expression being kind of confusing for students. I will do a version of it and see how it goes...

trevorcampbell: (By the way -- I am going to keep the original `filter` then `select` ordering from the R textbook, since this is fine as-is. If I wanted to write this from scratch I would indeed swap them, but I don't have much time left for this and am prioritizing important things.)

source/intro.md
trevorcampbell: @joelostblom Can you verify that it works in our python container please? That's where I tested it last night and it wasn't working.

(the image is listed in the `build_html.sh` script)

And then if it does, please edit the chapter directly to incorporate this info. I've added a little bold text where you should insert the new text if needed. 

If not, just delete that bolded bit please.

source/intro.md
trevorcampbell: handled

We should teach series and indexing in chapter 3. It is annoying that it isn't opt-in as you say, but I think we'd be hamstringing our students if we didn't at least go into it during wrangling.

source/intro.md
trevorcampbell: handled

source/intro.md
```diff
@@ -417,131 +418,212 @@ Wait a minute, nothing happened this time! Where's our data?
...
 And we can use that name to access the data frame and do things with it. 
-For example, we can type the name of the data frame to print the first few rows 
-on the screen. You will also see at the top that the number of observations (i.e., rows) and 
-variables (i.e., columns) are printed. Printing the first few rows of a data frame 
-like this is a handy way to get a quick sense for what is contained in a data frame.
+For example, we can type the name of the data frame to print both the first few rows 
+and the last few rows, separated by `...`,
+on the screen. You will also see at the bottom that the number of observations (i.e., rows) and 
+variables (i.e., columns) are printed. Printing a summary of a data frame 
+like this is a handy way to get a quick sense for what is contained in it.
```
joelostblom:
```suggestion
For example, we can type the name of the data frame to print both the first few rows 
and the last few rows. The three dots (`...`) indicate that there are additional rows that are not printed.
You will also see that the number of observations (i.e., rows) and 
variables (i.e., columns) are printed just underneath the data frame (214 rows and 6 columns in this case). 
Printing a few rows from data frame like this is a handy way to get a quick sense for what is contained in it.
```

source/intro.md
```diff
@@ -417,131 +418,212 @@ Wait a minute, nothing happened this time! Where's our data?
...
+## Creating subsets of data frames with `[]` & `loc[]`
 
-```{index} pandas.DataFrame; df[], pandas.DataFrame; loc[]
+```{index} pandas.DataFrame; [], pandas.DataFrame; loc[]
 ```
 
 Now that we've loaded our data into Python, we can start wrangling the data to
 find the ten Aboriginal languages that were most often reported
-in 2016 as mother tongues in Canada. In particular, we will construct 
+in 2016 as mother tongues in Canada. In particular, we want to construct
```
joelostblom, line 438:
```suggestion
Now that we've loaded our data into Python, we can start "wrangling" the data to
find the ten Aboriginal languages that were most often reported
in 2016 as mother tongues in Canada.
Data "wrangling" means that we are performing operations on the data
to get the result we want. Here, we want to construct
```

source/intro.md
```diff
@@ -577,7 +659,7 @@ The data are, therefore, in what we call a *tidy data* format. Tidy data is a
 fundamental concept and will be a significant focus in the remainder of this
 book: many of the functions from `pandas` require tidy data, including the
```
joelostblom:
```suggestion
book: many of the functions from `pandas` require tidy data, as does the
```

source/intro.md
```diff
@@ -600,16 +682,17 @@ import altair as alt
...
-The fundamental object in Altair is the `Chart`, which takes a data frame as a single argument `alt.Chart(ten_lang)`.
+The fundamental object in `altair` is the `Chart`, which takes a data frame as a single argument: `alt.Chart(ten_lang)`.
 With a chart object in hand, we can now specify how we would like the data to be visualized. 
-We first indicate what kind of geometric mark we want to use to represent the data. We can set the mark attribute 
-of the chart object using the `Chart.mark_*` methods.
-Here, as we want to plot the bar chart, so we will use `mark_bar()` method.
+We first indicate what kind of geometric mark we want to use to represent the data. Here we set the mark attribute 
+of the chart object using the `Chart.mark_bar` function, because we want to create a bar chart.
 Next, we need to encode the fields of the data frame using 
 the `x`(represents the x-axis position of the points) and
```
joelostblom:
```suggestion
Next, we need to encode the variables of the data frame using 
the `x` (represents the x-axis position of the points) and 
```

source/intro.md
```diff
@@ -600,16 +682,17 @@ import altair as alt
...
-of the chart object using the `Chart.mark_*` methods.
-Here, as we want to plot the bar chart, so we will use `mark_bar()` method.
+We first indicate what kind of geometric mark we want to use to represent the data. Here we set the mark attribute 
+of the chart object using the `Chart.mark_bar` function, because we want to create a bar chart.
 Next, we need to encode the fields of the data frame using 
 the `x`(represents the x-axis position of the points) and 
-`y`(represents the y-axis position of the points) channels. The `encode()`
-method builds a key-value mapping between encoding channels (such as x, y) 
-to fields in the dataset, accessed by field name(column names)
+`y`(represents the y-axis position of the points) *channels*. We use the `encode()`
```
joelostblom:
```suggestion
`y` (represents the y-axis position of the points) *channels*. We use the `encode()`
```

joelostblom (approved):
Looks great! I read it more in detail and had some minor linguistic suggestions. I didn't want to merge without you seeing them, so I will let you decide what to do with those and merge afterwards @trevorcampbell

source/intro.md
joelostblom: Ok I will look into that but I don't think that needs to hold up this PR. I opened a follow up issuehttps://github.com/UBC-DSCI/introduction-to-datascience-python/issues/69

source/intro.md
joelostblom: Great, thanks for considering and re-working this. I really like how it reads now and I think it works well with how you explained the boolean expression by just referring to it as a column instead of a series:

> We can use the equivalency operator == to compare the values of the category column with the value "Aboriginal languages"; you will learn about many other kinds of logical statements in Chapter [Cleaning and wrangling data](file:///home/joel/proj/mds/introduction-to-datascience-python/source/_build/html/wrangling.html#wrangling).

I do agree that it would be nice if we could use `[]` for everything, maybe that it just worked as `loc` if a comma was detected or something like that, not sure what the reasons are for why it doesn't. Maybe there would be conflict with some of the conveniences in `[]` versus the stricter behavior of `loc` and `iloc`.

joelostblom: The only thing I think could be confusing is that inside that boolean expression is a column selection with `[]` which we don't explain until the next section (`can_lang['category'] == 'Aboriginal languages'`), which I think we could solve by swapping the order of the section, but I saw your note about that you didn't want to prioritize that now which I also think makes sense.

PR #49, opened 2022-12-17
Production polish chapter 2

### LINKED ISSUE #4: reading
Progress:

Basic code has been converted from R to python. The conversion of text specific to R still needs to be converted to python. All the data files required for reading purpose have been pushed to the data folder. Have kept some R code still in the notebook and will take action on it after confirmation. The changes can be found on the `reading` branch of this repository in file `reading.md`

Issues and doubts regarding the `Reading` chapter:

- [x] `reading.py` not getting updated after building the `reading.md` file with `jb build reading.md`. 
- [x] What is the meaning of `\index{loading|see{reading}}\index{reading!definition}`, I think this is something specific to R markdown?
     - see https://jupyterbook.org/en/stable/content/content-blocks.html#indexes
- [x] Should the documentation only specific to R be deleted?
- [x] What to do regarding the absolute path, when reading `happiness_report.csv`
- [x] Should the column names be included while reading from `data/can_lang-meta-data.csv`
- [x] No need for `read_delim` dection, as it is already covered in the previous section using `sep=\t`, or should I just replace the documentation with - 'Any delimiter can be used with `sep=''`'
- [x] 3 ways of assigning column names- while reading(using `names=`), using `rename`, and using `df.columns`. which one to be used here? In the python tutorial, I saw `names` being used, but in `R`, `rename` is being used, so confused, which one to use
- [x] Installation of openpyxl, pgdb
- [x] What to do about `lazy evaluation` and `collect`, as python can read from Postres simply. should these be deleted from the documentation?
- [x] As the server `fakeserver.stat.ubc.ca` is an alias, the code in the python code-cell would error out, so should the code be in commented form?
- [x] To find the `min` of `average_rating`, can it be done in just one code cell? Unlike done in R?
- [x] What to do about additional resources section and exercises?
- [x] Confirm: I am reading images using the Myst syntax : 

```
{figure} ./path/to/figure.jpg
:name: label
caption
```
  - see https://jupyterbook.org/en/stable/content/figures.html#markdown-figures for a better approach

### LINKED ISSUE #36: Correct explanation of `read_table`
`read_table` is not a more flexible version of `read_csv`, it is the same function with a different default for the `sep` param. I think it is easier to just stick to `read_csv` and show how to change `sep`.

### LINKED ISSUE #38: Suggestions for Ch2 - Reading data
- [x] Add north/south vs gps corordinsdtea exmaples to illustrate relative and absolute paths
- [x] Don't emphasize "delimiter", it is referred to as "separator" in python.
- [x] Use text yntax highlighting instead of code for the preview of the file content
- [x] There are no "normal and expected messages" when reading in a file successfully with pandas
- [x] "how many lines to skip" -> "how many ROWS to skip"
- [x] Explain header = None
- [x] "So we needed to use different tools for the job" we use the same tool in Python
- [x] Don't use `inplace` with `rename` (or any pandas function, it is discouraged and will be deprecated)
    - [x] Also format this more properly
- [x] R - The learnobj says we should learn about `col_names` but we only ever use `rename`
- [x] Note potential issue with having to instal openxl separately for reading excel files with pandas
- [x] Update table at the end to remove `read_table`
- [x] We don't explain `autoload` and `autoload_with`
- [x] Why use head and tail first, instead of shape twice (more natural)?
- [x] Learning how to read and write records-formatted json files could be a valuable addition.
- [x] Fix formatting of pip install output
- [x] R & Py - Is the selector gadget better than built-in inspectors?
- [x] Format craigslist HTML with correct syntax highlighting
- [x] Pandas read_html page FileNotFoundError, explain "droplevels"
    - [x] I think we should also build up this method more to highlights its immense convenience, and not say that it is "fantastic" to read via beautiful soup.
- [x] I think scrapy is both more powerful and intuitive than beautiful soup
- [x] Twitter images broken
- [x] Introduce the print function and for loops before using them for tweepy
- [x] Another data file not found for the tweets
- [x] Is this SQL approach really easier than using `pd.read_sql`? For example ([more examples here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_sql.html#pandas.read_sql)):
    ```py
    import sqlalchemy as sqla
    db = sqla.create_engine("sqlite:///mydata.sqlite")
    pd.read_sql("SELECT * FROM test", db)
    ```
    - If we really want to remove all SQL syntax, we could use IBIS for a more pandas-like syntax, but I don't have experience with that personally https://ibis-project.org/docs/3.1.0/#features
- [x] #36

trevorcampbell:
~~Merge #46 before this one.~~ #46 has been merged, so this is safe now.

Done this now.

- closes #4 
- closes #38 
- closes #36


source/reading.md
```diff
@@ -15,13 +15,9 @@ kernelspec:
...
-We need to import the `pandas` package in order to read data into Python
-
-```{code-cell} ipython3
-import pandas as pd 
-```
 
 ```{code-cell} ipython3
+:tags: ["remove-cell"]
 import warnings
 warnings.filterwarnings('ignore')
```
joelostblom: Which warnings are we filtering with this? Could we resolve them instead of ignore them? Or at least set which specific warnings we are filtering so that we are not missing something important https://stackoverflow.com/questions/63979540/python-how-to-filter-specific-warning

source/reading.md
```diff
@@ -227,105 +219,103 @@ And here is a review of how we can use `read_csv` to load it into Python. First
...
 argument can be useful to tell Python 
-how many lines to skip before
+how many rows to skip before
 it should start reading in the data. In the example above, we would set this
-value to 2 and pass `header` as None to read and load the data correctly.
+value to 3 to read and load the data correctly.
 
 ```{code-cell} ipython3
-canlang_data = pd.read_csv("data/can_lang-meta-data.csv", skiprows=2, header=None)
+canlang_data = pd.read_csv("data/can_lang_meta-data.csv", skiprows=3, header=None)
```
joelostblom:
```suggestion
canlang_data = pd.read_csv("data/can_lang_meta-data.csv", skiprows=3)
```

We still want the column names to be detected:

image

source/reading.md
```diff
@@ -334,109 +324,107 @@ Non-Official & Non-Aboriginal languages Amharic 22465   12785   200 33670
...
+argument as the path to the file (as done with `read_csv`), and then provide
+values to the `sep` argument (here a tab, which we represent by `"\t"`),
+and finally set `header = None` to tell `pandas` that the data file does not
+contain its own column names.
 
 ```{code-cell} ipython3
-canlang_data =  pd.read_csv("data/can_lang.tsv", 
+canlang_data =  pd.read_csv("data/can_lang_no_cols.tsv", 
                            sep = "\t", 
                            header = None)
```
joelostblom:
```suggestion
canlang_data =  pd.read_csv(
    "data/can_lang_no_cols.tsv", 
    sep = "\t", 
    header = None
)
```

My suggestion is that we write like this in general; I think it is easier to read

source/reading.md
```diff
@@ -607,195 +592,186 @@ using Python with SQLite and PostgreSQL databases.
...
+`sqlite` backend in the
+`ibis` package. This command does not read
 in the data, but simply tells Python where the database is and opens up a
 communication channel that Python can use to send SQL commands to the database.
 
+> **Note:** There is another database package in python called `sqlalchemy`. 
+> That package is a bit more mature and popular than `ibis`,
+> so if you want to dig deeper into working with databases in Python, that is a good next 
+> package to learn about. We will work with `ibis` in this book, as it 
+> provides a friendlier syntax that is more like `pandas` for data analysis code.
```
joelostblom:
```suggestion
> provides a modern and friendlier syntax that is more like `pandas` for data analysis code. 
```

Just so that it doesn't sound like we are using some unpopular and not-preferred package

source/reading.md
```diff
@@ -607,195 +592,186 @@ using Python with SQLite and PostgreSQL databases.
...
-In order to actually retrieve this data in Python,
-we use the `fetchall()` function. The
-`sqlalchemy` package works this way because databases are often more efficient at selecting, filtering
-and joining large data sets than Python. And typically the database will not even
+Wait a second...this isn't the number of rows in the database. In fact, we haven't actually sent our 
+SQL query to the database yet! We need to explicitly tell `ibis` when we want to send the query.
+The reason for this is that databases are often more efficient at working with (i.e., selecting, filtering,
+joining, etc.) large data sets than Python. And typically, the database will not even
 be stored on your computer, but rather a more powerful machine somewhere on the
 web. So Python is lazy and waits to bring this data into memory until you explicitly
```
joelostblom:
```suggestion
web. So Ibis is lazy and waits to bring this data into memory until you explicitly
```
Maybe add a sentence of the benefits of this behavior? "Lazy" might otherwise sound like a bad thing

source/reading.md
```diff
@@ -808,126 +784,97 @@ Unlike SQLite,
...
-db = create_engine(connection_str)
-conn_mov_data = db.connect()
-
+```python
+conn = ibis.postgres.connect(database = "can_mov_db", 
+                                 host = "fakeserver.stat.ubc.ca",
+                                 port = 5432,
+                                 user = "user0001",
+                             password = "abc123"
+                            )
```
joelostblom:
```suggestion
conn = ibis.postgres.connect(
    database = "can_mov_db", 
    host = "fakeserver.stat.ubc.ca",
    port = 5432,
    user = "user0001",
    password = "abc123"
)
```

source/reading.md
```diff
@@ -808,126 +784,97 @@ Unlike SQLite,
...
 
 To find the lowest rating that exists in the data base, we first need to
-extract the `average_rating` column using `select`:
+select the `average_rating` column:
 
-```
-avg_rating_db = select([ratings.columns.average_rating])
-avg_rating_db
+```python
+avg_rating = ratings_table[["average_rating"]]
```
joelostblom, line 843: Are the double `[[` / `]]` needed here or is single enough?

source/reading.md
```diff
@@ -964,637 +909,637 @@ Databases are beneficial in a large-scale setting:
...
-languages data set without the official languages category according to the
+arguments are to use a comma (`,`) as the separator, and to include column names. 
+Below we demonstrate creating a new version of the Canadian
+languages data set without the "Official languages" category according to the
 Canadian 2016 Census, and then writing this to a `.csv` file:
 
 ```{code-cell} ipython3
-no_official_lang_data = canlang_data[canlang_data['category'] != 'Official languages']
+no_official_lang_data = canlang_data[canlang_data["category"] != "Official languages"]
 no_official_lang_data.to_csv("data/no_official_languages.csv")
```
joelostblom: I think we should recommend `index=False` because otherwise the index is written to the CSV, which is rarely desired and an odd default imo.

source/reading.md
```diff
@@ -964,637 +909,637 @@ Databases are beneficial in a large-scale setting:
...
+% `pandas.to_csv` to save it to a file and `pandas.read_csv` to read it into Python again later; 
+% and after reading the next few chapters you will have the skills to
+% compute the percentage of retweets versus tweets, find the most oft-retweeted
+% account, make visualizations of the data, and much more! If you decide that you want 
+% to ask the Twitter API for more data 
+% (see [the `tweepy` page](https://github.com/tweepy/tweepy)
+% for more examples of what is possible), just be mindful as usual about how much
+% data you are requesting and how frequently you are making requests.
+% 
+% +++
```
joelostblom, line 1534: Jupyter book should introduce block comments xD

joelostblom (approved):
Looks great overall, only minor comments and nothing that needs a second look, so I will approve and let you decide which comments to incorporate of the ones I made before merging.

source/reading.md
trevorcampbell: resolved by `ibis`

source/reading.md
trevorcampbell: resolved

source/reading.md
trevorcampbell: I'm sticking with double for consistency with Ch1, which only introduces selecting columns by putting a list between the sq bracks.

source/reading.md
trevorcampbell: ah agree -- I had this in mind when I was working but then forgot to actually do it ;) will fix

source/reading.md
trevorcampbell: I think this may have been a hold-over from Navya's original version (which did spit out a lot of warnings). Will check to see if those are all resolved now...

trevorcampbell: yup, there are no warnings. I will remove this.

source/reading.md
trevorcampbell: fixed

Issue #50, opened 2022-12-17
"Chapter [ref]" does not typeset well in python


trevorcampbell:
In the R version of the book, it typesets as "Chapter X" which is fine.

In python, it typesets as "Chapter [name of chapter]", e.g. "Chapter Python and Pandas", which doesn't read well.

We should just fix this by moving the word "Chapter" after the section reference. So now the text will read something like

"we will show in the Python and Pandas chapter, ..." (with a hyperlink)

trevorcampbell:
I will fix this in Chapters 1 and 2 as part of my polish pass before the semester in #48 and #49

but we should keep this issue open until we go through globally and fix

trevorcampbell:
~~Update -- I believe #44 changes how chapters are typeset to include numbers, so we will need to check whether our most recent edits break chapter links again and may need to revert back to the old style.~~

Scratch that, the original issue persists, so we should keep changing to the new wording.

trevorcampbell:
Right now we use `{ref}[reference name]`, but if we do `{numref}[reference name]` we should be able to go back to the original numerical chapter referencing.

However, we need to check how that typesets in PDF (and how the original `{ref}...` typesets too)

In HTML, doing `{numref}` typesets like "In Section X.Y, ..." -- "chapter" would be preferred over "section", although not for subsection refs. I wonder if we can configure that

trevorcampbell:
We can do 
```
{numref}`Chapter %s <ref-name-here>`
```

To typeset chapter refs in the same way as in the R book.

PR #76, opened 2022-12-23
Ch3: wrangling

### LINKED ISSUE #24: wrangling
The deployed book: https://gloriawyy.github.io/wrangling/wrangling.html
(hosted in this [repo](https://github.com/GloriaWYY/wrangling))

### LINKED ISSUE #39: Suggestions for Ch3 - Wrangling
- [x] `[]` instead of `df[]` in LOs
- [x] Use proper double newline after imports (pandas)
- [x] "In Python, pandas series are arrays with labels." I don't think we have defined "arrays" before this.
- [x] "to create the vector region as shown" -> "to create the SERIES region as shown"
- [ ] Fig 18 starts numbering at 1 for the series, but it starts at 0 in Python.
    - [ ] There is also no "character" type in Python, strings are read in as objects or strin
- [x] Table 3 focuses on general python types rather than pandas types. I think the latter makes more sense, so we should include `object` and `categorical` while removing `list`, `tuple`, `none` and `dict` (or include them in a separate table.
    - [x] We don't include all R types, e.g. matrices are excluded, so we don't need to include all Python types.
- [x] I think it makes more sense to talk about numpy arrays and/or dictionaries instead of "lists" when explaining pandas dataframes. (Named) lists are only appropriate for explaining R dataframes.
- [ ] Fig 19 shows R output with the boolean value in all caps
- [ ] R&Py I think it would be good with an explicit mention that the total size of the data can change when pivoting and specifically how it changes for this figure.
- [x] Explain `melt` more intuitively, use the word "melt" in the explanation, e.g. "We say that we "melt" (or "pivot") the wide table into a longer format".
- [x] We don't need to specify `value_vars` when we are using `id_vars` and want all the remaining columns to be melted. This would make the example code in the images significantly easier to read. (multiple places)
- [x] The note at the end of the `melt` section is incorrect similarly to one in the previous chapter. Python does not know to keep reading when lines end in commas, it is all about open parens. The comma is needed to separate the parameters, but it could come at any line (I believe this should be corrected in the R book too)
- [x] Figure 3.9 from the R book is missing
- [ ] Properly format the code in fig25 and fig 26 to have one parameter per row instead of being in one long line.
- [x] Why use dtypes here instead of `info` which we taught previously? (multiple places)
- [x] Explain that .str.split is using an accessors with functions/methods dedicated for a data type. Can be said in a simpler way but, should be introduced instead of just showing `str.split` as one thing.
- [x] `iloc` and `loc` are not needed here either, we can use ranges for columns with `[]`. (throughout this chapter)
    - [x] R&Py In general I think it is important to point out that if the order of the columns change unexpectingly, using a range can lead to unexpected output.
- [x] Just `[]` instead of `df[]`. I believe it is also not a function technically.
- [ ] I'm a bit hesitant to teaching `assign` before teaching regular assignment to a pandas column. The latter is often more straightforward, even if it can't be chained. I think `assign` might be better taught as advanced syntax and we would need to mentioned the `lambda` caveat when used in chains of a modified dataframe so that students don't easily do a mistake and use the unmodified dataframe to create a new column.
- [x] `query` is added in the chaining section without having been explained before.
- [x] R&Py I don't understand the note "You might also have noticed that we split the function calls across lines", isn't that the entire point of this section and it was just explained in the previous paragraph?
- [x] The section about chained indexing with `[]` followed by `loc` is discouraged and should be replaced with a single `loc` filtering both rows and columns (that's the main purpose of this indexer). (same under "more than two functions" section)
- [ ] Unnecessary whitespace in Fig 29
- [x] `by` keyword unnecessary for groupby
- [ ] reset_index not explained with agg, multi-index should at least be mentioned with a link to where to read more so that they know why we are resetting.
- [x] `iloc` slicing in agg section is unnecessary, just use `[]`
- [ ] `apply` is not needed for summary stats on multiple columns, the regular pandas functions work just fine and are much more performant and readable that the current approach with the vanilla python functions. `apply` is for custom functions.
- [ ] TODO I don't think skipna is needed here either
- [ ] The "Apply functions across many columns" is both incorrect and a repetition of the previous section.
- [x] lambda function used without introduction
- [ ] We don't need apply to apply across one row either, again the built-in functions are preferred with axis='row' or 1.
- [ ] I think this last section chaining is quite messy overall and could use a complete rewrite of the code. We should make sure that all the functionality from the R book is included here since the languages diverge in their approaches on these topics.
    - apply could be its own section talking about using functions that are not in pandas, but being discouraged for everything that is built into pandas already.

---

Some TODOs for myself to look more into to see if these sections can be improved as well:

- TODO Check the pivoting under fig 29, seems odd to reassign columns like that
- TODO important lacking functionality from pivot_table
- TODO simplify code under fig 27 to not use pd.concat (or at least introduce concat before this)? This could also simplyfy the `to_numeric` section. Are we doing method chaining anywhere else?
- TODO - [ ] df.filter + a regex instead of the startswith approach? At least easier for "contains", need to know ^ for startswith
- TODO `as_dtype` instead of `to_numeric`
- TODO double check NA handling in pandas min max, or why are we using python mix max? that is why na is inclueded... pandas would not. The entire assign aggregation section is odd, who assigns an empty df like this? Same for groupby

lheagy:
Another pass at the wrangling chapter. I have summarized comments / questions in #39. This is still going to need some work. I would appreciate input on the comments I made in #39. Feel free to push directly to this branch too! 

This PR:
- closes #39 
- closes #24


lheagy:
I took another crack at this! I moved the discussion of chaining to the intro chapter and have re-organized so that `assign` is at the very end. We are not worried about always ending up with a data frame at the end of each operation, so this simplifies the code. It is a pretty substantial re-write, so another set of eyes on the flow would be very much appreciated!

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
 only two columns presents some difficulty: for example, you have to scrutinize
-the table quite closely to get a sense for the relative numbers of speakers of 
-each language. When you move on to more complicated analyses, this issue only 
-gets worse. In contrast, a *visualization* would convey this information in a much 
-more easily understood format. 
+the table quite closely to get a sense for the relative numbers of speakers of
+each language. When you move on to more complicated analyses, this issue only
+gets worse. In contrast, a *visualization* would convey this information in a much
+more easily understood format.
 Visualizations are a great tool for summarizing information to help you
```
trevorcampbell: this text is no longer valid here. need a better transition to chaining and move this argument regarding viz to right before the viz section.

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
-effectively communicate with your audience. 
+effectively communicate with your audience.
+
+
+## Combining functions by chaining the methods
+
+```{index} chaining methods
+```
+
+In Python, we often have to call multiple methods in a sequence to process a data
```
trevorcampbell: could be a smoother transition here and flow better from previous text. Can reference the fact that we need three ops (loc, sort, head), so what is the right way to do these?

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
+many steps. Let's revisit the question: What are the 10 most commonly spoken Aboriginal languages?
+
+Starting from the `can_lang` data frame, we solved this in multiple steps:
+
+1) Filter the rows so that we only have the category `Aboriginal languages`
+2) Select the colums `language` and `mother_tongue` (Note that this can be combined with step 1 using `loc`)
+3) Sort the values by `mother_tongue` with the largest values at the top
+4) Ask for only the top 10 values using `head`
+
+One way of performing these three steps is to just write
```
trevorcampbell: looks like four steps in the list currently, but this says 3

also we should order these steps in the same way they were presented in this section (not sure if currently order is the same)

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
+multiple lines of code, storing temporary objects as you go:
```
trevorcampbell: publisher asked us to avoid putting ":" before code blocks, if I recall. Can just end the sentence with a period

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
+```{code-cell} ipython3
+aboriginal_lang = can_lang.loc[can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]]
+arranged_lang_sorted = aboriginal_lang.sort_values(by='mother_tongue', ascending=False)
+ten_lang = arranged_lang_sorted.head(10)
+```
+This requires creating multiple temporary, intermediate outputs, and can get quite long.
+
+An alternative approach uses "chaining" to link multiple operations together. This can
```
trevorcampbell: unnecessary paragraph break

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
+4) Ask for only the top 10 values using `head`
+
+One way of performing these three steps is to just write
+multiple lines of code, storing temporary objects as you go:
+```{code-cell} ipython3
+aboriginal_lang = can_lang.loc[can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]]
+arranged_lang_sorted = aboriginal_lang.sort_values(by='mother_tongue', ascending=False)
+ten_lang = arranged_lang_sorted.head(10)
+```
+This requires creating multiple temporary, intermediate outputs, and can get quite long.
```
trevorcampbell: - hard to read ("quite long" is saying this only implicitly)
- suggests that the named objects are of some importance, when really they are just intermediates

source/intro.md
```diff
@@ -631,20 +632,70 @@ We have now answered our initial question by generating this table!
...
+
+An alternative approach uses "chaining" to link multiple operations together. This can
+simplify the code and make it easier to read. The code below accomplishes the same task:
+```{code-cell} ipython3
+# obtain the 10 most common Aboriginal languages
+ten_lang = (
+    can_lang.loc[can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]]
+    .sort_values(by="mother_tongue", ascending=False)
+    .head(10)
+    )
```
trevorcampbell: This block of code needs way more explanation. Like really break it down. What do the dots mean? why can I keep using dots over and over again? Why do I put each on its own line? is it another expression? etc etc etc

source/wrangling.md
```diff
@@ -55,57 +55,19 @@ By the end of the chapter, readers will be able to do the following:
       - `.str.split`
   - Recall and use the following operators for their
     intended data wrangling tasks:
```
trevorcampbell, line 57: check the learning objectives for validity.
- "define what lists, series, and data frames are" -- do we do lists explicitly now, e.g.?

source/wrangling.md
```diff
@@ -55,57 +55,19 @@ By the end of the chapter, readers will be able to do the following:
...
-#       - `!`
-#       - `&` 
-#       - `|`
-#       - `|>` and `%>%`
-```
+      - `.iloc[]`
 
 ## Data frames, series, and lists
 
 In Chapters {ref}`intro` and {ref}`reading`, *data frames* were the focus:
```
trevorcampbell: The code
```
In Chapter {ref}`blah`, we do...
```
does not typeset properly in jupyterbook. See #50 . Need to search through whole chapter and change to

```
the {ref}`blah` chapter
```

source/wrangling.md
```diff
@@ -152,45 +114,29 @@ data set. There are 13 entities in the data set in total, corresponding to the
...
 
-```{code-cell} ipython3
-:tags: [remove-cell]
-
-# The following cell was removed because there is no "vector" in Python.
-```
-
-+++ {"tags": ["remove-cell"]}
-
-Python stores the columns of a data frame as either
```
trevorcampbell, line 163: We need to keep some amount of this text (rewritten appropriately) to say that dataframes use series as columns, otherwise students won't know why you are covering series below

source/wrangling.md
```diff
@@ -152,45 +114,29 @@ data set. There are 13 entities in the data set in total, corresponding to the
...
 
 ```{index} pandas.Series
 ```
 
-In Python, `pandas` **series** are arrays with labels. They are strictly 1-dimensional and can contain any data type (integers, strings, floats, etc), including a mix of them (objects);
-Python has several different basic data types, as shown in {numref}`tab:datatype-table`.
-You can create a `pandas` series using the `pd.Series()` function.  For 
-example, to create the vector `region` as shown in
-{numref}`fig:02-series`, you can write:
+In Python, `pandas` **series** are lists. They are ordered and can be indexed.
```
trevorcampbell: "series are lists"

I kind of prefer the R wording here: "are objects that can contain one or more elements"

also series technically not a python list

maybe write the R phrasing and then say something like "(essentially, a list)"

source/wrangling.md
```diff
@@ -152,45 +114,29 @@ data set. There are 13 entities in the data set in total, corresponding to the
...
+They are strictly 1-dimensional and can contain any data type
```
trevorcampbell: "1-dimensional" will not mean anything to a first year. less jargon

source/wrangling.md
```diff
@@ -152,45 +114,29 @@ data set. There are 13 entities in the data set in total, corresponding to the
...
+(integers, strings, floats, etc), including a mix of them; Python
```
trevorcampbell: although it can include a mix, we should make sure to remark that it is generally good practice to store only one kind of data in a series (because all observations of one variable should be the same type)

source/wrangling.md
```diff
@@ -152,45 +114,29 @@ data set. There are 13 entities in the data set in total, corresponding to the
...
+has several different basic data types, as shown in
+{numref}`tab:datatype-table`.
+You can create a `pandas` series using the
+`pd.Series()` function.  For example, to create the series `region` as shown
+in{numref}`fig:02-series`, you can write:
```
trevorcampbell: spacing error

source/wrangling.md
trevorcampbell: also reinforce here that series are what pandas uses for each column of a df

source/wrangling.md
```diff
@@ -265,76 +176,26 @@ image_read("img/data_frame_slides_cdn/data_frame_slides_cdn.007.jpeg") %>%
...
 ```
 
 +++
 
-It is important in Python to make sure you represent your data with the correct type. 
-Many of the `pandas` functions we use in this book treat 
+It is important in Python to make sure you represent your data with the correct type.
+Many of the `pandas` functions we use in this book treat
 the various data types differently. You should use integers and float types
 (which both fall under the "numeric" umbrella type) to represent numbers and perform
```
trevorcampbell: "numeric" isn't a type in python -- this was missed from R translation

source/wrangling.md
```diff
@@ -265,76 +176,26 @@ image_read("img/data_frame_slides_cdn/data_frame_slides_cdn.007.jpeg") %>%
...
 arithmetic. Strings are used to represent data that should
-be thought of as "text", such as words, names, paths, URLs, and more. 
+be thought of as "text", such as words, names, paths, URLs, and more.
 There are other basic data types in Python, such as *set*
```
trevorcampbell: We need to discuss each type that appears in the table.

Missing bool, nonetype: what are these used for?

In the text here we should also reference the type name, so `int` `float` `bool` `str` and `NoneType` after we mention them in the sentence

E.g. 

"Strings (`str`) blahblah"

source/wrangling.md
```diff
@@ -265,76 +176,26 @@ image_read("img/data_frame_slides_cdn/data_frame_slides_cdn.007.jpeg") %>%
...
-| dictionary            | `dict`     | Mapping Type   | mapping of key-value pairs                    | `{'name':'DSCI', 'code':100, 'credits':2}` |
-| none                  | `NoneType` | Null Object    | represents no value                           | `None`                                     |
+| Data type             | Abbreviation | Description                                   | Example                                    |
+| :-------------------- | :----------- | :-------------------------------------------- | :----------------------------------------- |
+| integer               | `int`        | positive/negative whole numbers               | `42`                                       |
+| floating point number | `float`      | real number in decimal form                   | `3.14159`                                  |
+| boolean               | `bool`       | true or false                                 | `True`                                     |
+| string                | `str`        | text                                          | `"Can I have a cheezburger?"`              |
```
trevorcampbell: "Can I have a cheezburger" is cute but kind of out-of-place informality here

source/wrangling.md
```diff
@@ -345,10 +206,10 @@ A vector versus a list.
 
 A data frame is really just series stuck together that follows two rules:
```
trevorcampbell: "is really just series stuck together"

series is its own plural here, so it's a bit awk

maybe

"A data frame is really just a list of series....."

source/wrangling.md
```diff
@@ -345,10 +206,10 @@ A vector versus a list.
...
 
-1. Each element itself is a series. 
+1. Each element itself is a series.
 2. Each element (series) must have the same length.
 
-Not all columns in a data frame need to be of the same type. 
+Not all columns in a data frame need to be of the same type.
 {numref}`fig:02-dataframe` shows a data frame where
 the columns are series of different types.
```
trevorcampbell: Should take some of the text from the R version to put here at the end. Like

"But each element in one column should usually be the same type, since the values for a single variable are usually all of the same type. So even though series let you put different types in them, most common to have just one per column blahblah"

source/wrangling.md
```diff
@@ -386,49 +230,29 @@ Data frame and vector types.
 > For example we can check the class of the Canadian languages data set,
 > `can_lang`, we worked with in the previous chapters and we see it is a `pandas.core.frame.DataFrame`.
 
-```{code-cell} ipython3
-:tags: [remove-cell]
-
-# The functions from the `tidyverse` package that we use often give us a
-# special class of data frame called a *tibble*. Tibbles have some additional \index{tibble}
```
trevorcampbell, line 393: We should keep this whole paragraph, but reword it to be about `pandas DataFrames`

There will be no mention of a "base python dataframe" because that doesn't exist here. But it's still worth saying what the type is, discussing that dataframes provide a bunch of functions (like loc and [] and whatever).

Try to keep the R pedagogy as much as possible

source/wrangling.md
```diff
@@ -386,49 +230,29 @@ Data frame and vector types.
 > For example we can check the class of the Canadian languages data set,
```
trevorcampbell, line 239: the code cell below this note is not inside the note. It should be inside the note (see the R version)

source/wrangling.md
```diff
@@ -386,49 +230,29 @@ Data frame and vector types.
...
-# > For example we can check the class of the Canadian languages data set,
-# > `can_lang`, we worked with in the previous chapters and we see it is a tibble.
-```
 
 ```{code-cell} ipython3
 can_lang = pd.read_csv("data/can_lang.csv")
 type(can_lang)
 ```
 
 Lists, Series and DataFrames are basic types of *data structure* in Python, which
```
trevorcampbell: The table link here is broken (it points elsewhere in the chapter...)

also, should we cover `list`, `dict`, `tuple` and `set` here? This would probably be the place to do it. Also say "while this course will focus on data frames in Pandas, python has some very useful data structures that you will often see"

Table:
Data structure |  Example |   Usage
list   | [1,"a", 4] | storing an ordered collection of elements. You can index it like `blah`.
dict | {a : 3, blah} | storing a mapping from key to value

etc etc

source/wrangling.md
```diff
@@ -491,18 +315,18 @@ below!
...
-is to combine values that are stored in separate columns, 
+One task that is commonly performed to get data into a tidy format
+is to combine values that are stored in separate columns,
 but are really part of the same variable, into one.
-Data is often stored this way 
-because this format is sometimes more intuitive for human readability 
+Data is often stored this way
+because this format is sometimes more intuitive for human readability
 and understanding, and humans create data sets.
-In {numref}`fig:02-wide-to-long`,
```
trevorcampbell, line 500: The caption on the figure that this reference is pointing to is overlapping the image itself somehow (Fig 17). No idea why, but figure shouldn't overlap its caption

source/wrangling.md
```diff
@@ -518,13 +342,16 @@ greatly simplified once the data is tidied.
...
+population size? Land area? It's not clear.
+To solve both of these problems,
+we can reshape this data set to a tidy data format
 by creating a column called "year" and a column called
 "population." This transformation&mdash;which makes the data
 "longer"&mdash;is shown as the right table in
-{numref}`fig:02-wide-to-long`.
+{numref}`fig:02-wide-to-long`. Note that the number of entries in our data frame
+can change in this transformation. The "untidy" data has 5 rows and 3 columns for
+a total of 15 data, whereas the "tidy" data on the right has 15 rows and 2 columns
```
trevorcampbell: We don't usually refer to entries in a data frame in this class as "data" (even though that's technically correct). I would use "entries" here

source/wrangling.md
```diff
@@ -541,41 +368,42 @@ Melting data from a wide to long data format.
 ```
 
 We can achieve this effect in Python using the `.melt` function from the `pandas` package.
-The `.melt` function combines columns, 
-and is usually used during tidying data 
-when we need to make the data frame longer and narrower. 
+We say that we "melt" (or "pivot") the wide table into a longer format.
```
trevorcampbell: I would remove this new "We say that" sentence -- especially because "pivot" is actually a different function in python (the R naming is so much better in this case...pivot_wider and pivot_longer. Python has "melt" and "pivot").

source/wrangling.md
```diff
@@ -613,13 +441,11 @@ We use `.melt` to combine the Toronto, Montréal,
 Vancouver, Calgary, and Edmonton columns into a single column called `region`,
 and create a column called `mother_tongue` that contains the count of how many
 Canadians report each language as their mother tongue for each metropolitan
-area. We specify `value_vars` to be all
-the columns between Toronto and Edmonton:
+area:
 
 ```{code-cell} ipython3
 lang_mother_tidy = lang_wide.melt(
     id_vars=["category", "language"],
```
trevorcampbell, line 509: This code block is different from the image above

Need to add a bold black comment for melissa to update the image (if the current code block is what we want to go with)

source/wrangling.md
```diff
@@ -613,13 +441,11 @@ We use `.melt` to combine the Toronto, Montréal,
 Vancouver, Calgary, and Edmonton columns into a single column called `region`,
 and create a column called `mother_tongue` that contains the count of how many
 Canadians report each language as their mother tongue for each metropolitan
-area. We specify `value_vars` to be all
-the columns between Toronto and Edmonton:
+area:
```
trevorcampbell: avoid colon before code block

source/wrangling.md
```diff
@@ -754,7 +581,7 @@ lang_home_tidy
 ```
 
 ```{code-cell} ipython3
-lang_home_tidy.dtypes
+lang_home_tidy.info()
```
trevorcampbell, line 669: holy crap this is a useful function!

We should absolutely teach `.info()` earlier. How about in a new section right before "Tidy data" (i.e., at the end of the "Data frames, series, and lists" section)

trevorcampbell: (and of course explain all the printouts you get with `.info()`

source/wrangling.md
```diff
@@ -55,57 +55,19 @@ By the end of the chapter, readers will be able to do the following:
...
-#       - `==` 
-#       - `%in%`
-#       - `!`
-#       - `&` 
-#       - `|`
-#       - `|>` and `%>%`
-```
+      - `.iloc[]`
 
 ## Data frames, series, and lists
```
trevorcampbell: Do we cover lists any more? rename this section?

source/wrangling.md
```diff
@@ -754,7 +581,7 @@ lang_home_tidy
 ```
```
trevorcampbell, line 647: Just above this: `lang_home_tidy.columns = [....]` is not explained at all. What is going on there? Do we need to do that any time we pivot? 

(also if we're renaming columns, we taught students how to do that in Ch 1 -- we should follow that example for consistency)

trevorcampbell: `reset_index` is also not explained at all

source/wrangling.md
```diff
@@ -795,28 +622,34 @@ lang_messy
...
 lang_messy_longer
 ```
 
-Next we'll use `.str.split` to split the `value` column into two columns. 
-One column will contain only the counts of Canadians 
-that speak each language most at home, 
-and the other will contain the counts of Canadians 
-that speak each language most at work for each region. 
+Next we'll use `.str.split` to split the `value` column into two columns.
+How it works is that it takes a single string and splits it into multiple values
```
trevorcampbell: awk wording

source/wrangling.md
```diff
@@ -795,28 +622,34 @@ lang_messy
...
+based on the character you tell it to split on. For example:
```
trevorcampbell: avoid colon before code block

source/wrangling.md
```diff
@@ -843,7 +676,7 @@ tidy_lang
 ```
```
trevorcampbell, line 781: again, lots of unexplained stuff going on above. `rename` and `drop`, `concat`, etc.

If we can simplify this code, we should

If we cannot (or can only to a small extent), we need to explain everything that is going on here in far more detail

source/wrangling.md
```diff
@@ -843,7 +676,7 @@ tidy_lang
...
 
 ```{code-cell} ipython3
-tidy_lang.dtypes
+tidy_lang.info()
```
trevorcampbell: Do we need to print info here? or is the table print above enough?

source/wrangling.md
trevorcampbell: It's often a good idea to "look at the output" of a partial operation, go "oh man there's something wrong now!" and then fix it. This is way less mystical for students, because they see individually why each step is happening.

trevorcampbell (changes requested):
I reviewed Ch1, and Ch3 up to "Using [] to extract rows or columns" (not including)

I figure I suggested enough changes that I will hold off on reviewing the rest until the current batch is sorted.

source/intro.md
lheagy: :+1: will combine to 3. The order is the same if you start from "Using `loc[]` to filter rows and select columns"

source/wrangling.md
lheagy: I think the concept of a "list of series" might be a bit less intuitive than "stuck together" I will try a less backward wording, but if this doesn't fit with you, I am happy to adopt your suggestion

"
A data frame is really just a collection of series that are stuck together; it follows two rules
"

source/wrangling.md
lheagy: I took a pass at this. Rather than including examples in the table, I included code examples below. See if you think this works

source/wrangling.md
lheagy: I think it might be best to stick with the simpler code, so I will add a note

source/intro.md
```diff
@@ -619,32 +620,88 @@ arranged_lang
...
+```
+
+What just happened? The code above starts with parentheses `()` and is split over multiple lines.
```
lheagy: added this paragraph here. See if you think this does the trick!

source/wrangling.md
lheagy: I broke this up into a few steps starting at 750. See if you think that is an improvement

source/wrangling.md
```diff
@@ -2005,138 +1349,259 @@ An alternative to aggregating on a dataframe
...
+    value_name="value",
+)
+tidy_lang_obj = (
+    pd.concat(
+        (lang_messy_longer, lang_messy_longer["value"].str.split("/", expand=True)),
+        axis=1,
+    )
+    .rename(columns={0: "most_at_home", 1: "most_at_work"})
+    .drop(columns=["value"])
+)
```
lheagy: I am not sure this is critical to repeat. We could just pick up with `official_langs`? 

But I can also appreciate that it might be useful to revisit and see it all together

source/wrangling.md
lheagy: updated the content, so now we explicitly do lists and dictionaries

source/wrangling.md
```diff
@@ -386,48 +242,74 @@ Data frame and vector types.
...
 ```{code-cell} ipython3
 can_lang = pd.read_csv("data/can_lang.csv")
 type(can_lang)
 ```
 
-Lists, Series and DataFrames are basic types of *data structure* in Python, which
-are core to most data analyses. We summarize them in 
-{numref}`tab:datastructure-table`. There are several other data structures in the Python programming 
-language (*e.g.,* matrices), but these are beyond the scope of this book.
+### Data structures in Python
```
lheagy, line 248: This may not need its own title, but I found it helpful. Feel free to remove if you think it is best to flow from the content about DataFrames above

source/wrangling.md
```diff
@@ -265,76 +180,29 @@ image_read("img/data_frame_slides_cdn/data_frame_slides_cdn.007.jpeg") %>%
...
 the various data types differently. You should use integers and float types
-(which both fall under the "numeric" umbrella type) to represent numbers and perform
-arithmetic. Strings are used to represent data that should
-be thought of as "text", such as words, names, paths, URLs, and more. 
+to represent numbers and perform arithmetic. Integers have no decimals while floats contain decimals.
+Booleans are variables that can only take on one of two values: True or False.
+Strings are used to represent data that should
+be thought of as "text", such as words, names, paths, URLs, and more.
+A `NoneType` is a special type in Python that is used to indicate no-value; this can occur when
+you have missing data.
```
lheagy: added this per comment below

source/wrangling.md
lheagy: See new section on `### Data structures in Python` (starting on 251)

source/wrangling.md
lheagy: I have no idea why. I started #87 . It is still readable, so I don't think critical to solve at the moment

source/wrangling.md
lheagy: added the text below. See what you think!

source/wrangling.md
lheagy: added lines 655 and following to explain this. It doesn't seem like it needs its own heading? but we can easily add a heading if you prefer -- no strong opinions on my end

source/wrangling.md
lheagy: hmm, that doesn't easily work in juypterbook... I created #88

source/wrangling.md
```diff
@@ -152,45 +114,32 @@ data set. There are 13 entities in the data set in total, corresponding to the
 A data frame storing data regarding the population of various regions in Canada. In this example data frame, the row that corresponds to the observation for the city of Vancouver is colored yellow, and the column that corresponds to the population variable is colored blue.
```
joelostblom, line 114: It would be nice if this image also highlighted what a value is, for both R and Py

image

source/wrangling.md
```diff
@@ -386,46 +238,72 @@ Data frame and vector types.
...
+| ---            | ----------- |
+| list | An ordered collection of values that can store multiple data types at once. |
+| dict | A labeled data structure where `keys` are paired with `values` |
+| Series | An ordered collection of values *with labels* that can store multiple data types at once. |
+| DataFrame | A labeled data structure with `Series` columns of potentially different types. |
+```
+
+A `list` is an ordered collection of values. To create a list, we put the contents of the list in between
+square brackets `[]`, where each item of the list is separated by a comma. A `list` can contain values
+of different types. The example below contains 6 `str` entries.
```
joelostblom:
```suggestion
of different types. The example below contains six `str` entries.
```

source/wrangling.md
```diff
@@ -386,46 +238,72 @@ Data frame and vector types.
...
 ```
 
-Lists, Series and DataFrames are basic types of *data structure* in Python, which
-are core to most data analyses. We summarize them in 
-{numref}`tab:datastructure-table`. There are several other data structures in the Python programming 
-language (*e.g.,* matrices), but these are beyond the scope of this book.
+### Data structures in Python
 
-+++
+The `Series` and `DataFrame` types are *data structure* in Python, which
```
joelostblom:
```suggestion
The `Series` and `DataFrame` types are *data structures* in Python, which
```

source/wrangling.md
joelostblom: I think this title is helpful so I would vote for keeping it.

source/wrangling.md
```diff
@@ -437,15 +315,15 @@ language (*e.g.,* matrices), but these are beyond the scope of this book.
 
 There are many ways a tabular data set can be organized. This chapter will focus
 on introducing the **tidy data** format of organization and how to make your raw
```
joelostblom:
```suggestion
There are many ways a tabular data set can be organized. The data frames we have looked at so far
have all been using the **tidy data** format. This chapter will focus on introducing this type of
data frame format of organization and walk through how to make your raw
```


I think the start of this tidy data section is repetition of the "What is a data frame" section where we already introduced the concept of rows, columns, and variables. Maybe we can keep only one of these illustrations and start this section off by referring to the previous one as in my suggestion above?

source/wrangling.md
joelostblom: I think it would be nice if we used the [jupyter book admonitions](https://jupyterbook.org/en/stable/content/content-blocks.html#notes-warnings-and-other-admonitions) ("note" in this case) instead of block quotes for these section. These can also contain code as demonstrated in the link.

source/wrangling.md
```diff
@@ -541,41 +424,41 @@ Melting data from a wide to long data format.
 ```
 
 We can achieve this effect in Python using the `.melt` function from the `pandas` package.
```
joelostblom: I think we decided to just use `method` instead of `.method`, so we should remove the prefixed `.` here and elsewhere

source/wrangling.md
```diff
@@ -629,9 +512,10 @@ lang_mother_tidy
...
 > `.melt` function is split across several lines. This is allowed in
-> certain cases; for example, when calling a function as above, as long as the 
-> line ends with a comma `,` Python knows to keep reading on the next line.
-> Splitting long lines like this across multiple lines is encouraged 
+> certain cases; for example, when calling a function as above, the input
+> arguments are between parentheses `()` and Python knows to keep reading on
+> the next line. Each line ends with a comma `,` making it easier to read.
+> Splitting long lines like this across multiple lines is encouraged
 > as it helps significantly with code readability. Generally speaking, you should
 > limit each line of code to about 80 characters.
```
joelostblom: Aren't we already introducing method chaining in the intro section chapter? In that case we can remove this from here I think.

source/wrangling.md
```diff
@@ -698,14 +582,14 @@ lang_long = pd.read_csv("data/region_lang_top5_cities_long.csv")
...
+However, each observation is split across multiple rows:
+one where the count for `most_at_home` is recorded,
+and the other where the count for `most_at_work` is recorded.
+Suppose the goal with this data was to
 visualize the relationship between the number of
-Canadians reporting their primary language at home and work. 
+Canadians reporting their primary language at home and work.
 Doing that would be difficult with this data in its current form,
 since these two variables are stored in the same column.
 {numref}`fig:img-pivot-wider-table` shows how this data
```
joelostblom, line 600: It would be nice if this figure showed the params on separate lines as with the code block below and the image for `melt`

joelostblom: Also, it is unnecessary to specify `values` here since all remaining columns will be used for the values parameter. This is the same reason we omitted `value_vars` above, so I think it would make sense to omit it here too for consistency and simplicity. I think we can add a sentence mentioning that these parameter exists for both sections

source/wrangling.md
```diff
@@ -753,8 +639,27 @@ lang_home_tidy.columns = [
...
+In the first step, note that we added a call to `reset_index`. When `pivot` is called with
+multiple column names passed to the `index`, those entries become the "name" of each row that
+would be used when you filter rows with `[]` or `loc` rather than just simple numbers. This
+can be confusing... What `reset_index` does is sets us back with the usual expected behaviour
+where each row is "named" with an integer. This is a subtle point, but the main take-away is that
+when you call `pivot`, it is a good idea to call `reset_index` afterwards.
+
+The second operation we applied is to rename the columns. When we perform the `pivot`
+operation, it keeps the original column name `"count"` and adds the `"type"` as a second column name.
+Having two names for a column can be confusing! So we rename giving each column only one name.
```
joelostblom, line 658: I think it won't be clear to students why we do this when we previously used `rename`. I don't think there is a better way to flatten an index, [most other approaches ](https://datascientyst.com/flatten-multiindex-in-pandas/)would take more explaining. But it might make sense to explicitly say what we are doing here since it is different from what we taught them previously

source/wrangling.md
joelostblom: I don't think this needs its own heading but agree that we should introduce it early. One possible spot is when we talk about data frame column types and the `type` function. We mention the importance of having the same type in each every column there, but we do'n't show them how to see which data types are in a column, which `info` would remedy.

source/wrangling.md
joelostblom: An alternative to concat + rename is to assign directly to two columns in the dataframe:

```py
lang_messy_longer[["most_at_home", "most_at_work"]] = lang_messy_longer["value"].str.split("/", expand=True)
```

I think this is a bit easier to read myself (after showing them the output from the right hand expression once first), but if we want to make a point of showing concat because we think it is generally useful then I think the current way looks good.

source/wrangling.md
```diff
@@ -795,57 +701,88 @@ lang_messy
...
+we will contatenate (combine) the original data frame with `split_counts` using
+the `concat` function from `pandas`. The `concat` function *concatenates* data frames
+along an axis. By default, it concatenates the data frames vertically along `axis=0` yielding a single
+*taller* data frame. Since we want to concatenate our old columns to our
+new `split_counts` data frame (to obtain a *wider* data frame), we will specify `axis=1`.
+
+```{code-cell} ipython3
+tidy_lang = pd.concat(
+        (lang_messy_longer, split_counts),
         axis=1,
```
joelostblom:
```suggestion
tidy_lang = pd.concat(
    [lang_messy_longer, split_counts],
    axis=1,
```

Same indentation as elsewhere and change to a list because we don't teach tuples I think?

source/wrangling.md
```diff
@@ -854,175 +791,65 @@ Is this data set now tidy? If we recall the three criteria for tidy data:
 
 We can see that this data now satisfies all three criteria, making it easier to
 analyze. But we aren't done yet! Notice in the table, all of the variables are
```
joelostblom:
```suggestion
analyze. But we aren't done yet! Although we can't see it in the data frame above, all of the variables are actually
```

The dtype is not shown in pandas as it is in dplyr

source/wrangling.md
```diff
@@ -854,175 +791,65 @@ Is this data set now tidy? If we recall the three criteria for tidy data:
...
+This change is due to the delimiter (`/`) when we read in this messy data set.
+Python read these columns in as string types, and by default, `str.split` will
+return columns with the `object` data type.
+
+It makes sense for `region`, `category`, and `language` to be stored as an
+`object` type. However, suppose we want to apply any functions that treat the
+`most_at_home` and `most_at_work` columns as a number (e.g., finding rows
+above a numeric threshold of a column).
+That won't be possible if the variable is stored as a `object`.
+Fortunately, the `pandas.to_numeric` function provides a natural way to fix problems
```
joelostblom, line 821: There is also `infer_objects` and `convert_dtypes` which should do this automatically (but for some reason doesn't work here it seems... might still be worth a mention)

source/wrangling.md
```diff
@@ -1124,25 +951,6 @@ pd.Series(["Vancouver", "Toronto"]) == pd.Series(["Toronto", "Vancouver"])
...
-# > to all the elements in `vectorB`, and so on. Notice the difference between `==` and
-# > `%in%` in the example below.
-# > 
-# >``` {r}
-# >c("Vancouver", "Toronto") == c("Toronto", "Vancouver")
-# >c("Vancouver", "Toronto") %in% c("Toronto", "Vancouver")
-# >```
-```
-
 ### Extracting rows above or below a threshold using `>` and `<`
```
joelostblom, line 972: Should we also mention `>=` and <=`?

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
-only English in Toronto is reported by more people 
-as their primary language at home 
+This operation returns a data frame with only one row, indicating that when
+considering the official languages,
+only English in Toronto is reported by more people
+as their primary language at home
 than French in Montréal according to the 2016 Canadian census.
 
-+++ {"tags": []}
+### Extracting rows using `query`
```
joelostblom, line 1002: I wonder if it will be confusing that we are introducing a second way of doing the same thing... maybe we should add an explicit comment on that the other way is the default way? Or remove `query` unless we make heavy use of it later in the book

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+official_langs.query("most_at_home > 2669195")
+```
 
-### Using `.assign` to modify columns
+The query (criteria we are using to select values) is input as a string. This
+comes in handy when we chain operations together.
 
-```{index} pandas.DataFrame; df[]
+(loc-iloc)=
+## Using `loc[]` to filter rows and select columns.
```
joelostblom, line 1017: To me it seems like everything that is done is this section could be done with just `[]` and there is no need to involve `loc` here. We can save `loc` for only those situation when we are both filtering rows and selecting columns, similar to what we did in the intro chapter.

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
-  geom_bar(stat = "identity") +
-  xlab("Region") +
-  ylab("Proportion of Canadians that speak English most often at home")
-```
-
-From this visualization, we can see that in Calgary, Edmonton, Toronto, and
-Vancouver, English was reported as the most common primary language used at
-home.  However, in Montréal, this does not seem to be the case!
--->
+## Aggregating data with `agg`
```
joelostblom:
```suggestion
## Aggregating data
```

Since we cover more than just `agg` in this section

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+> see an input variable `skipna`, which by default is set to `skipna=True`. This means that
+> `pandas` will skip `NaN` values when computing statistics.
+
+### Calculating summary statistics on dataframes
 
-```{index} pandas.DataFrame; sort_values
+What if you want to calculate summary statistics on an entire dataframe? Well,
+it turns out that the functions in {numref}`tab:basic-summary-statistics`
+can be applied to a whole data frame!
+For example, we can ask for the number of rows that each column has using `count`.
```
joelostblom, line 1254: Hmmm, is this a useful thing to show? Doesn't data frames require that all column have the same number of rows so we would use `info` or `shape` to get the number of rows in the data frame instead?

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+```{code-cell} ipython3
+region_lang.count()
+```
+Not surprisingly, they are all the same. We could also ask for the `mean`, but
+some of the columns in `region_lang` contain string data with words like `"Vancouver"`
+and `"Halifax"`---for these columns there is no way for `pandas` to compute the mean.
+So we provide the keyword `numeric_only=True` so that it only computes the mean of columns with numeric values. This
```
joelostblom, line 1261: How annoying that pandas is starting to require this, I quite liked the convenient of not having to specify it... it seems like a reasonable default to only take the average of numerical columns... I wonder what their rationale is.

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+```{code-cell} ipython3
+region_lang.min()
 ```
 
-As we saw in Chapter {ref}`intro`, 
-we can use the `.sort_values` function 
-to order the rows in the data frame by the values of one or more columns. 
-Here we pass the column name `most_at_home` to sort the data frame rows by the values in that column, in ascending order.
+Similarly, if there are only some columns that you would like to get summary statistics of,
+we can first use `loc[]` and then ask for the summary statistic. An example of this is illustrated in {numref}`fig:summarize-across`.
```
joelostblom:
```suggestion
we can first use `[]` and then ask for the summary statistic. An example of this is illustrated in {numref}`fig:summarize-across`.
```

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
-large_region_lang
+`loc[]` or `apply` is useful for efficiently calculating summary statistics on many columns at once. The darker, top row of each table represents the column headers.
 ```
 
+Lets say that we want to know
+the mean and standard deviation of all of the columns between `"mother_tongue"` and `"lang_known"`.
+We use `loc[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
 ```{code-cell} ipython3
-:tags: [remove-cell]
+region_lang.loc[:, "mother_tongue":"lang_known"].agg(["mean", "std"])
```
joelostblom, line 1295:
```suggestion
region_lang["mother_tongue":"lang_known"].agg(["mean", "std"])
```

If we make the change I suggestion to this section above and only use loc for both filtering rows and selecting coluns at the same time

source/wrangling.md
```diff
@@ -1152,1262 +960,890 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+:name: fig:summarize-across
+:figclass: caption-hack
 
-large_region_lang
+`loc[]` or `apply` is useful for efficiently calculating summary statistics on many columns at once. The darker, top row of each table represents the column headers.
 ```
 
+Lets say that we want to know
+the mean and standard deviation of all of the columns between `"mother_tongue"` and `"lang_known"`.
+We use `loc[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
```
joelostblom, line 1293:
```suggestion
We use `[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
```

joelostblom (changes requested):
This is looking great! I have provided a few suggestions, most of which are minor. I will go through the rest of the PR tomorrow.

source/wrangling.md
```diff
@@ -1152,1262 +974,807 @@ glue("census_popn", "{0:,.0f}".format(35151728))
...
+city_populations
 ```
-
-We see that we get an additional column added to the data frame, 
-named `maximum`, which is the maximum value between `mother_tongue`,
-`most_at_home`, `most_at_work` and `lang_known` for each language
-and region.
-
+This new data frame has the same `region` column as the `english_lang` data frame. The order of
+the cities is different, but that is okay, we can use the `merge` function in `pandas` to say
```
lheagy: note that I added the introduction of `merge` here. It seems like a natural example but was not in the previous version, so please see what you think

source/wrangling.md
lheagy: :+1: I like your suggestion @joelostblom: please feel free to change

source/wrangling.md
trevorcampbell: I think it's fine (and often good) to repeat things like this -- but maybe we should link back to the explanation in Ch1?

source/wrangling.md
trevorcampbell: agree!

source/wrangling.md
trevorcampbell: yes

trevorcampbell: but no need for a special section (or title) for it. Just mention it in this section

source/wrangling.md
lheagy: :+1: see the section on Data Structures in Python (starting at line 247)

source/wrangling.md
trevorcampbell: more wrangling more better!

source/wrangling.md
lheagy: Note added in 481

Issue #56, opened 2022-12-17
Exercises links


trevorcampbell:
The exercises link at the end of each chapter (as well as the chapter title where necessary) has to be fixed 

I've created a new empty repo in the `ubc-dsci` org for the python worksheets

https://github.com/UBC-DSCI/data-science-a-first-intro-python-worksheets

Please link to this using the code
```
[worksheets repository](https://github.com/UBC-DSCI/data-science-a-first-intro-python-worksheets#readme)
```

in the exercises sections

PR #77, opened 2022-12-25
Ch4: Viz

### LINKED ISSUE #25: vizualisation
1. may consult the Dates and Times chapter of *R for Data Science* [@wickham2016r];
What to use as a python substitute?
2. @joelostblom   While creating the plot for filtered dates, I am filtering the dates using `.loc[]` operator and then plotting using the following code:
```
co2_df = pd.read_csv("mauna_loa_data.csv", parse_dates=['date_measured'])
co2_dates = co2_df.loc[(co2_df.date_measured >= '1990-01-01') &  (co2_df.date_measured < '1993-01-01')]
co2_dates
co2_line = alt.Chart(co2_dates).mark_line(color='black').encode(
    x = alt.X("date_measured:T",title = "Year"),
    y = alt.Y("ppm", scale=alt.Scale(zero=False), title = "Atmospheric CO2 (ppm)")).configure_axis(
    titleFontSize=12)
```
The output of this plot is as below with X-axis labels not coming year wise, but a mixture of year and month

image

Desired output:
image

### LINKED ISSUE #40: Suggestions for Ch4 - Viz

- [x] LOs much shorter than R chapter
- [ ] R&Py Soften section about avoiding pie charts and 3D
- [ ] Add more specific comment on Vega colorschemes
- [ ] R&Py Consider moving the checklists from "Refining the visualization" to the end, this is a lot of info that we are front-loading learners with, most of it likely not meaningful until they have seen some viz. It is nice with a few words on the purpose of viz, but this might be too much details too early.
- [ ] R&Py Fix this sentence to be more correct, zooming in on small difference is completely ok when the difference is meaningful (e.g. rise in earth's temperature) "Don’t adjust the axes to zoom in on small differences. If the difference is small, show that it’s small!"
    - [ ] Especially since ggplot zoom in by default as we can see in figure 4.3 and 4.4
- [x] "To use the altair package, we need to import the altair package. We will also import pandas in order to support reading and other data related operations." -> "We will also import pandas to use for reading in the data."
- [x] .info instead of .dtypes
- [x] Incomplete sentence "while using the altair package, We create a plot object"
- [ ] Remove note about how to make it look like ggplot "Note: We can change the size of the point and color of the plot by specifying mark_point(size=10, color='black')."
- [x] Why do we set `color='black'` in the lineplot with no explanation?
- [x] Dangling end parenthesis would be more in line with the syntax used in the altair docs.
- [x] Axis is being zoomed without any explanation as to why `scale(zero=false)`, both line and scatter.
- [x] Much of the code use spaces around `=` in parameter assignment which is R syntax, not Python.
- [x] Fig 37 uses long lines of code that are hard to read
    - [ ] (we should really reset figure numbering per chapter and number chapters/parts)
- [x] This note is not very encouraging to students and could be expressed in another way "Note: The configure_ function in altair is complex and supports many other functionalities, which can be viewed here" (i usually don't teach configure at all)
- [x] `Scale(zero=false)` is explained now after being used multiple times already
- [x] The code for fig 38 is really hard to read, use more newlines.
- [ ] Why are we setting the tickcount in fig 38?
- [ ] R&Py instead of zooming in on the figure, why don't we filter the data? Is there a specific advantage of zooming in that we want students to learn here?
- [x] Fig 39, again the visualization is styled to look like the one in the R textbook instead of going with the default in Altair; this adds unnecessary noise to the code and makes it harder to learn.
- [x] Do we want to talk about `filled=True` and `mark_circle` here?
- [x] Fig 42 & 43 & 44 code needs more newlines (just go through all figures)
- [x] Incorrect newline in "ongue language has 19460850 speakers, \nwhile the least common"
- [x] `.assign` indentation is used incorrectly
- [x] The legend is incorrectly described as taking up the plot area. This is a limitation of ggplot but not of altair which have a constant area for the chart and the legend is added outside of that.
- [x] The section about color palette has improper paragraph delineation.
- [x] We should mention that the default color palette in altair is appropriate for most situations since it is the same as was developed by researchers at Tableau. This is again different from ggplot which uses evenly distributed color around a color wheel which is really inappropriate and should never be used as it includes red and green to separate categorical values.
- [x] https://www.color-blindness.com/coblis-color-blindness-simulator/ certificate expired
- [ ] Considering limiting ouput to 6 row with hidden pandas config?
- [x] `.sort_values(by = "size", ascending=False).iloc[:12]` -> `nlargest(12)`
- [x] "our question ("**WHICH** are the top "
- [ ] Is it really necessary to use `configure` axis on each chart? It seems like this takes away from the key concepts we are learning and adds noise.
- [x] fig50 `mark_bar(color='black')` and then also use the color encoding?
- [x] "convert the morley data to a tibble to take " Remove `tibble`
- [ ] R Histograms can become more visually appealing by setting the color outline to 'white'
- [ ] Why is the section on "histograms" using a bar chart without binning?
- [ ] The vertical line in altair is currently drawn using a separate dataframe instead of just alt.value/alt.datum
- [ ] Need to explain altair data types here since we are starting to use `:N`
- [ ] I don't think we explained what `+` does?
- [x] Weird indentation for `.assign` again
- [x] The default number of bins in altair is not 30. It uses scott's or silverman's rules to compute it and then snaps it (down) to evenly align with the d3 axis grid.
- [x] "helping **TO** answer our question"
- [ ] Need a more elaborate section on saving files and common issues
- [ ] Zoomed in image from R
- [ ] Add additional resources
- [ ] R In an introductory course like this, it could be beneficial to teach `geom_col` instead of `geom_bar(stat = 'identity')`, which simplifies the explanation and reduces potential confusion. `geom_bar` would then only be used for counts. The drawback is that most people in the R community seem to not be aware of `geom_col` so you often see the more cryptic `geom_bar(stat = 'identity') in the wild (but the tidyverse recommends it https://ggplot2.tidyverse.org/reference/geom_bar). We also use `geom_histogram` instead of `geom_bar` with `stat_bin` or `scale_x_bin`.
- [ ] R Why not use `facet_wrap(~continent, ncol=1)` instead of `facet_grid(rows = vars(continent))`? It seems both more intuitive and common to me, and it is also the recommended approach in the tidyverse docs https://ggplot2.tidyverse.org/reference/facet_grid.html

lheagy:
First pass at addressing issues in #40. Lots of formatting updates and work to simplify the code 

This PR:
- closes #40 
- closes #25


source/viz.md
```diff
@@ -15,23 +15,27 @@ kernelspec:
...
+    - bar plots
+    - histogram plots
 - Given a data set and a question, select from the above plot types and use Python to create a visualization that best answers the question.
 - Given a visualization and a question, evaluate the effectiveness of the visualization and suggest improvements to better answer the question.
 - Referring to the visualization, communicate the conclusions in non-technical terms.
-- Identify rules of thumb for creating effective visualizations. 
+- Identify rules of thumb for creating effective visualizations.
 - Define the two key aspects of altair objects:
     - mark objects
     - encodings
```
trevorcampbell, line 54: Below here there's a `{-}` in a section heading -- I assume that's something from the R version?

source/viz.md
```diff
@@ -58,22 +62,22 @@ Imagine your visualization as part of a poster presentation for a project; even
...
 As with most coding tasks, it is totally fine (and quite common) to make
 mistakes and iterate a few times before you find the right visualization for
 your data and question. There are many different kinds of plotting
-graphics available to use (see Chapter 5 of *Fundamentals of Data Visualization* {cite:p}`wilkeviz` for a directory). 
+graphics available to use (see Chapter 5 of *Fundamentals of Data Visualization* {cite:p}`wilkeviz` for a directory).
 The types of plot that we introduce in this book are shown in {numref}`plot_sketches`
-which one you should select depends on your data 
-and the question you want to answer. 
-In general, the guiding principles of when to use each type of plot 
+which one you should select depends on your data
```
trevorcampbell, line 91: missing semicolon in this sentence after figure ref -- see R version

source/viz.md
trevorcampbell: I actually see a lot of these throughout the chapter, should probably remove them -- I dont think that syntax is used in jupyterbook

source/viz.md
```diff
@@ -126,17 +130,17 @@ understand and remember your message quickly.
 - Ensure the text, symbols, lines, etc., on your visualization are big enough to be easily read.
 - Ensure the data are clearly visible; don't hide the shape/distribution of the data behind other objects (e.g.,  a bar).
 - Make sure to use color schemes that are understandable by those with
-  colorblindness (a surprisingly large fraction of the overall 
+  colorblindness (a surprisingly large fraction of the overall
   population&mdash;from about 1% to 10%, depending on sex and ancestry {cite:p}`deebblind`).
-  For example, [Color Schemes](https://vega.github.io/vega/docs/schemes/) 
+  For example, [Color Schemes](https://vega.github.io/vega/docs/schemes/)
```
trevorcampbell: Is this actually a useful link for `altair` ? We should make sure whatever we point the students to here is useful for their actual work

source/viz.md
```diff
@@ -172,74 +176,65 @@ from myst_nb import glue
...
-> even though they contain non-numeric characters 
+> For example, `datetime` type vectors allow functions like `altair`
+> to treat them as numeric dates and not as character vectors,
+> even though they contain non-numeric characters
 > (e.g., in the `date_measured` column in the `co2_df` data frame).
-> This means Python will not accidentally plot the dates in the wrong order 
-> (i.e., not alphanumerically as would happen if it was a character vector). 
+> This means Python will not accidentally plot the dates in the wrong order
+> (i.e., not alphanumerically as would happen if it was a character vector).
 > More about dates and times can be viewed [here](https://wesmckinney.com/book/time-series.html)
```
trevorcampbell: Missing period at the end of this sentence

source/viz.md
```diff
@@ -172,74 +176,65 @@ from myst_nb import glue
...
 ```
 
-We see that there are two columns in the `co2_df` data frame; `date_measured` and `ppm`. 
-The `date_measured` column holds the date the measurement was taken, 
+We see that there are two columns in the `co2_df` data frame; `date_measured` and `ppm`.
+The `date_measured` column holds the date the measurement was taken,
 and is of type `datetime64`.
-The `ppm` column holds the value of CO$_{\text{2}}$ in parts per million 
+The `ppm` column holds the value of CO$_{\text{2}}$ in parts per million
 that was measured on each date, and is type `float64`.
```
trevorcampbell: "and is type `float64`" -- remind students that this is just the usual type for  decimal numbers

source/viz.md
```diff
@@ -253,7 +248,7 @@ There are a few basic aspects of a plot that we need to specify:
 - The **geometric encoding**, which tells `altair` how the columns in the data frame map to properties of the visualization.
     - To create an encoding, we use the `encode()` function.
     - The `encode()` method builds a key-value mapping between encoding channels (such as x, y) to fields in the dataset, accessed by field name(column names)
```
trevorcampbell: "field name(column names)" -- needs a space before parens

source/viz.md
```diff
@@ -262,41 +257,37 @@ from myst_nb import glue
 
 ```{code-cell} ipython3
 co2_scatter = alt.Chart(co2_df).mark_point().encode(
-    x = "date_measured", 
-    y = alt.Y("ppm", scale=alt.Scale(zero=False)))
-    
-
+    x="date_measured",
+    y=alt.Y("ppm", scale=alt.Scale(zero=False))
```
trevorcampbell, line 295: We need to explain the `scale=alt.Scale...` in the text before its first usage.

Make sure to remove the explanation that occurs below around its second usage.

source/viz.md
```diff
@@ -262,41 +257,37 @@ from myst_nb import glue
...
-:figwidth: 700px 
+:::{glue:figure} co2_scatter
+:figwidth: 700px
 :name: co2_scatter
 
 Scatter plot of atmospheric concentration of CO$_{2}$ over time.
 :::
 
-
-> **Note:** We can change the size of the point and color of the plot by specifying `mark_point(size=10, color='black')`.
```
trevorcampbell, line 284: This code example isn't useful unless we show students what it does -- need to use it and explain it

Also why is it here? It seems to be about `mark_point`, but we've moved on to `mark_line`. Probably move elsewhere

trevorcampbell: (A good place to put this is if we have a scatter example later that gets refined -- would fit well in that refinement)

trevorcampbell: Woops, sorry -- I added this comment to the wrong line in the code file. It still pertains to where the note shows up later in the chapter.

source/viz.md
```diff
@@ -368,42 +369,47 @@ glue('co2_line_labels', co2_line_labels, display=False)
...
 :::{glue:figure} co2_line_labels
-:figwidth: 700px 
+:figwidth: 700px
 :name: co2_line_labels
 
 Line plot of atmospheric concentration of CO$_{2}$ over time with clearer axes and labels.
 :::
 
-> **Note:** The `configure_` function in `altair` is complex and supports many other functionalities, which can be viewed [here](https://altair-viz.github.io/user_guide/configuration.html)
+> **Note:** The `configure_` function in `altair` supports many other functionalities, which can be viewed [here](https://altair-viz.github.io/user_guide/configuration.html)
```
trevorcampbell: Give more detail. What are these other functionalities? 

Also `configure_` isn't a function, it's a collection of functions and probably should be referred to with a star, so like

"There are many `configure_*` functions in `altair`, which..."

source/viz.md
```diff
@@ -412,31 +418,31 @@ glue('co2_line_scale', co2_line_scale, display=False)
...
 
 Line plot of atmospheric concentration of CO$_{2}$ from 1990 to 1994.
 :::
 
-Interesting! It seems that each year, the atmospheric CO$_{\text{2}}$ increases until it reaches its peak somewhere around April, decreases until around late September, 
+Interesting! It seems that each year, the atmospheric CO$_{\text{2}}$ increases until it reaches its peak somewhere around April, decreases until around late September,
 and finally increases again until the end of the year. In Hawaii, there are two seasons: summer from May through October, and winter from November through April.
 Therefore, the oscillating pattern in CO$_{\text{2}}$ matches up fairly closely with the two seasons.
 
 As you might have noticed from the code used to create the final visualization
```
trevorcampbell: This "as you might have noticed" para flows poorly now (and odd para break below)

trevorcampbell: (see R version for original text)

source/viz.md
```diff
@@ -502,20 +509,17 @@ We can see in {numref}`faithful_scatter` that the data tend to fall
 into two groups: one with short waiting and eruption times, and one with long
 waiting and eruption times. Note that in this case, there is no overplotting:
 the points are generally nicely visually separated, and the pattern they form
-is clear.  Also, note that to make the points solid, we used `filled=True` as argument of the `mark_point` function. In place of `mark_point(filled=True)`, we can also use `mark_circle()`. 
+is clear.  Also, note that to make the points solid, we used `filled=True` as argument of the `mark_point` function. In place of `mark_point(filled=True)`, we can also use `mark_circle()`.
```
trevorcampbell: repeated comment here (keep the first instance, and move the remark about mark_circle there)

source/viz.md
```diff
@@ -644,9 +653,9 @@ much more readable and interpretable now. However, the scatter points themselves
...
 up in the lower left-hand side of the visualization. The data is clumped because
 many more people in Canada speak English or French (the two points in
-the upper right corner) than other languages. 
-In particular, the most common mother tongue language 
-has {glue:}`numlang_speakers_max` speakers,  
+the upper right corner) than other languages.
+In particular, the most common mother tongue language
+has {glue:}`numlang_speakers_max` speakers,
 while the least common has only {glue:}`numlang_speakers_min`.
 That's a {glue:}`log_result` -decimal-place difference
```
trevorcampbell: It's definitely not a 14 decimal place difference

I suspect log base e is being used here, when it should be log base 10

also this should be typeset using integers, not floats

source/viz.md
```diff
@@ -716,16 +736,16 @@ glue("result", result)
 
 ```
 
-Similar to some of the examples in Chapter {ref}`wrangling`, 
-we can convert the counts to percentages to give them context 
+Similar to some of the examples in Chapter {ref}`wrangling`,
```
trevorcampbell: "In Chapter {ref}wrangling" won't typeset properly. See #50 and earlier comments on Ch3 PR

source/viz.md
```diff
@@ -848,37 +889,50 @@ glue('can_lang_plot_category', can_lang_plot_category, display=False)
...
+:figwidth: 700px
 :name: can_lang_plot_category
 
 Scatter plot of percentage of Canadians reporting a language as their mother tongue vs the primary language at home colored by language category.
 :::
 
 
 The legend in {numref}`can_lang_plot_category`
-takes up valuable plot area. 
+adds a fair bit of horizontal space.
```
trevorcampbell: I'm not sure this story works well in the python version. If you look in the R version, the figure gets *super* squished. In the python version, honestly the side legend vs the top legend both look about the same to me.

We should fix this at some point, but if you don't want to handle it now you can open a separate issue for it and punt it for now.

source/viz.md
```diff
@@ -887,31 +941,31 @@ glue('can_lang_plot_legend', can_lang_plot_legend, display=False)
...
 :name: can_lang_plot_legend
 
 Scatter plot of percentage of Canadians reporting a language as their mother tongue vs the primary language at home colored by language category with the legend edited.
 :::
 
 In {numref}`can_lang_plot_legend`, the points are colored with
-the default `altair` color palette. But what if you want to use different
-colors? In Altair, there are many themes available, which can be viewed [here](https://vega.github.io/vega/docs/schemes/)
+the default `altair` color palette. This is an appropriate choice for most situations. If you want to use different
+colors? In Altair, there are many themes available, which can be viewed [in the documentation](https://vega.github.io/vega/docs/schemes/)
```
trevorcampbell: The text here flows poorly now, and lots of weird para breaks

source/viz.md
```diff
@@ -920,16 +974,30 @@ further improve the clarity and accessibility of your visualization.
 
 ```{code-cell} ipython3
```
trevorcampbell, line 1020: The note box right above this code doesn't need to be a note box. Just write in the text that we are using (switching back to??) mark point and specifying shape attr

source/viz.md
```diff
@@ -1000,13 +1068,13 @@ groups of a categorical variable.
 ```
 
 We specify that we would like to use a bar plot
-via the `mark_bar` function in `altair`. 
+via the `mark_bar` function in `altair`.
 The result is shown in {numref}`islands_bar`
```
trevorcampbell: Missing period

source/viz.md
```diff
@@ -1031,14 +1099,15 @@ the `sort_values` function followed by the `iloc` property.  Then to help us mak
...
 
-```{index} pandas.DataFrame; sort_values, pandas.DataFrame; iloc[]
+```{index} pandas.DataFrame; nlargest
 ```
 
 ```{code-cell} ipython3
-islands_top12 = islands_df.sort_values(by = "size", ascending=False).iloc[:12] 
+islands_top12 = islands_df.nlargest(12)
 
 islands_bar_sorted = alt.Chart(islands_top12).mark_bar().encode(
```
trevorcampbell: This code doesn't work -- stack trace printed

source/viz.md
```diff
@@ -1031,14 +1099,15 @@ the `sort_values` function followed by the `iloc` property.  Then to help us mak
 space, we'll use horizontal bars instead of vertical ones. We do this by
 swapping the `x` and `y` variables:
```
trevorcampbell: avoid colon before code block (please fix this everywhere in this chapter and open an issue for it for a global pass)

source/viz.md
```diff
@@ -1047,30 +1116,28 @@ glue('islands_bar_sorted', islands_bar_sorted, display=True)
...
-To organize the landmasses by their `size` variable, 
+To organize the landmasses by their `size` variable,
 we will use the `altair` `sort` function
 in encoding for `y` axis to organize the landmasses by their `size` variable, which is encoded on the x-axis.
 To sort the landmasses by their size(denoted on `x` axis), we use `sort='x'`. This plots the values on `y` axis
-in the ascending order of `x` axis values. 
+in the ascending order of `x` axis values.
 
 We do this here so that the largest bar will be closest to the axis line,
 which is more visually appealing.
```
trevorcampbell: Note box below here is not necessary. Just put it as part of the text that you can do descending by `-x`

source/viz.md
```diff
@@ -1047,30 +1116,28 @@ glue('islands_bar_sorted', islands_bar_sorted, display=True)
...
 
-To organize the landmasses by their `size` variable, 
+To organize the landmasses by their `size` variable,
 we will use the `altair` `sort` function
 in encoding for `y` axis to organize the landmasses by their `size` variable, which is encoded on the x-axis.
 To sort the landmasses by their size(denoted on `x` axis), we use `sort='x'`. This plots the values on `y` axis
-in the ascending order of `x` axis values. 
+in the ascending order of `x` axis values.
 
 We do this here so that the largest bar will be closest to the axis line,
```
trevorcampbell, line 1189: Weird para break and sentence by itself

source/viz.md
```diff
@@ -1086,15 +1153,15 @@ The default label is the name of the column being mapped to `color`. Here that
...
+Finally, we again use the `configure_axis` function
 to change the font size.
 
 ```{code-cell} ipython3
-islands_plot_sorted = alt.Chart(islands_top12).mark_bar(color='black').encode(
-    x = alt.X("size",title = "Size (1000 square mi)"),
-    y = alt.Y("landmass", title = "Landmass", sort='x'),
-    color = alt.Color("landmass_type", title = "Type")).configure_axis(
-    titleFontSize=12)
+islands_plot_sorted = alt.Chart(islands_top12).mark_bar().encode(
```
trevorcampbell, line 1204: Code does not work, stack trace

source/viz.md
```diff
@@ -1191,76 +1257,75 @@ Histogram of Michelson's speed of light data.
...
+We would also like to fine tune this vertical line,
 styling it so that it is dashed and 1 point in thickness.
-A point is a measurement unit commonly used with fonts, 
-and 1 point is about 0.353 mm. 
-We do this by setting `strokeDash=[3,3]` and `size = 1`, respectively. 
+A point is a measurement unit commonly used with fonts,
+and 1 point is about 0.353 mm.
+We do this by setting `strokeDash=[3,3]` and `size = 1`, respectively.
 
 Similarly, a horizontal line can be plotted using the `y` axis encoding and the dataframe with one value, which would act as the be the y-intercept
```
trevorcampbell: Lots of weird para breaks here again

source/viz.md
```diff
@@ -1191,76 +1257,75 @@ Histogram of Michelson's speed of light data.
...
 
-Note that 
-*vertical lines* are used to denote quantities on the *horizontal axis*, 
-while *horizontal lines* are used to denote quantities on the *vertical axis*. 
+Note that
+*vertical lines* are used to denote quantities on the *horizontal axis*,
+while *horizontal lines* are used to denote quantities on the *vertical axis*.
 
 To add the dashed line on top of the histogram, we will use the `+` operator. This concept is also known as layering in altair.(This is covered in the later sections of the chapter). Here, we add the `mark_rule` chart on the `morley_hist` chart of the form `mark_bar`
```
trevorcampbell: Poor formatting of parenthetical

Also, *is* it covered later? If not, we should probably cover it here where it is first used

trevorcampbell: honestly we should probably just cover it here anyway, because it's being used here

source/viz.md
```diff
@@ -1281,27 +1346,28 @@ Histogram of Michelson's speed of light data colored by experiment.
 ```{index} integer
 ```
 
-Alright great, {numref}`final_plot_colored` looks...wait a second! We are not able to distinguish 
-between different Experiments in the histogram! What is going on here? Well, if you 
+Alright great, {numref}`final_plot_colored` looks...wait a second! We are not able to distinguish
+between different Experiments in the histogram! What is going on here? Well, if you
 recall from Chapter {ref}`wrangling`, the *data type* you use for each variable
```
trevorcampbell: Chapter {ref}wrangling again -- see #50

source/viz.md
```diff
@@ -1281,27 +1346,28 @@ Histogram of Michelson's speed of light data colored by experiment.
...
 with the data types in the `morley` data frame. In particular, the `Expt` column
 is currently an *integer*. But we want to treat it as a
-*category*, i.e., there should be one category per type of experiment.  
+*category*, i.e., there should be one category per type of experiment.
 
 ```{index} nominal, altair; :N
 ```
 
-To fix this issue we can convert the `Expt` variable into a `nominal`(categorical) type 
+To fix this issue we can convert the `Expt` variable into a `nominal`(categorical) type
```
trevorcampbell: bad formatting with parens (needs a space before)

please check through whole chapter for this and fix

source/viz.md
```diff
@@ -1281,27 +1346,28 @@ Histogram of Michelson's speed of light data colored by experiment.
...
 and the color will be mapped discretely. Here, we also mention `stack=False`, so that the bars are not stacked on top of each other.
 
 ```{code-cell} ipython3
 morley_hist_categorical = alt.Chart(morley_df).mark_bar(opacity=0.5).encode(
-    x = alt.X("Speed", bin=alt.Bin(maxbins=50)),  
-    y=alt.Y('count()', stack=False),
-    color = "Expt:N")
+    x=alt.X("Speed", bin=alt.Bin(maxbins=50)),
+    y=alt.Y("count()", stack=False),
+    color="Expt:N"
```
trevorcampbell: This `:N` thing is neat -- is it the standard way of accomplishing this in altair? seems a bit too "magical" to me in comparison with other things, but I'm no expert so I'm not sure.

source/viz.md
```diff
@@ -1333,25 +1399,27 @@ grid of separate histogram plots.
...
-morley_hist = alt.Chart(morley_df).mark_bar(opacity = 0.5).encode(
-    x = alt.X("Speed", bin=alt.Bin(maxbins=50)),  
-    y=alt.Y('count()', stack=False),
-    color = "Expt:N").properties(height=100, width=300)
-
-final_plot_facet = (morley_hist + v_line).facet(row = 'Expt:N', data = morley_df)
+morley_hist = alt.Chart(morley_df).mark_bar(opacity=0.5).encode(
+    x=alt.X("Speed", bin=alt.Bin(maxbins=50)),
+    y=alt.Y("count()", stack=False),
+    color="Expt:N"
```
trevorcampbell: This code does not produce the intended plot. The plot is now faceted, but all the facets have the same (multicolor) histogram. Please refer to R version of textbook to see what this should look like.

source/viz.md
```diff
@@ -1379,75 +1447,94 @@ The most different experiments still obtained quite similar results!
 ```
 
 There are two finishing touches to make this visualization even clearer. First and foremost, we need to add informative axis labels
-using the `alt.X` and `alt.Y` function, and increase the font size to make it readable using the `configure_axis` function. Second, and perhaps more subtly, even though it 
-is easy to compare the experiments on this plot to one another, it is hard to get a sense 
+using the `alt.X` and `alt.Y` function, and increase the font size to make it readable using the `configure_axis` function. Second, and perhaps more subtly, even though it
+is easy to compare the experiments on this plot to one another, it is hard to get a sense
 of just how accurate all the experiments were overall. For example, how accurate is the value 800 on the plot, relative to the true speed of light?
 To answer this question, we'll use the assign function to transform our data into a relative measure of accuracy rather than absolute measurements:
```
trevorcampbell: assign should have code tags

source/viz.md
trevorcampbell: see the following plot -- somehow that one is right

source/viz.md
```diff
@@ -1379,75 +1447,94 @@ The most different experiments still obtained quite similar results!
...
 #### Choosing a binwidth for histograms
 
-When you create a histogram in `altair`, the default number of bins used is 30.
+When you create a histogram in `altair`, by default, it tries to choose a reasonable  number of bins.
 Naturally, this is not always the right number to use.
 You can set the number of bins yourself by using
 the `maxbins` argument in the `mark_bar` geometric object.
 
-But what number of bins is the right one to use? 
+But what number of bins is the right one to use?
```
trevorcampbell, line 1620: more weird para breaks

source/viz.md
```diff
@@ -1457,80 +1544,126 @@ On the other hand, the `maxbins=200` and `maxbins=5` are too small and too big,
...
 
 ```{code-cell} ipython3
 :tags: ["remove-cell"]
-glue('final_plot_max_bins', final_plot_max_bins, display=True)
+glue("final_plot_max_bins", final_plot_max_bins, display=True)
 ```
 
 :::{glue:figure} final_plot_max_bins
-:figwidth: 700px 
+:figwidth: 700px
```
trevorcampbell, line 1727: This figure appears to be missing -- it wasn't rendered in the HTML

source/viz.md
```diff
@@ -1457,80 +1544,126 @@ On the other hand, the `maxbins=200` and `maxbins=5` are too small and too big,
...
 :name: final_plot_max_bins
 
 Effect of varying number of max bins on histograms.
 :::
 
-#### Adding layers to a `altair` plot object {-}
+#### Adding layers to a `altair` plot object {+}
```
trevorcampbell: this subsec now has a `{+}` after it -- should be removed, unless there is some special reason this syntax is needed

trevorcampbell: Ah I see what happened. Note that `{-}` is a hold over from the R version (it isn't meant to be the subtraction sign for layering plots or whatever, it's special syntax for R bookdown)

The `{+}` can just be removed here

source/viz.md
```diff
@@ -1457,80 +1544,126 @@ On the other hand, the `maxbins=200` and `maxbins=5` are too small and too big,
...
 
 ```{index} altair; +
 ```
 
-One of the powerful features of `altair` is that you 
+One of the powerful features of `altair` is that you
```
trevorcampbell: OK yeah, regarding my previous point about layering: yes, we should just cover it earlier where vertical line is used first. We can remove this subsection on layering, it won't be necessary once we cover it earlier.

source/viz.md
```diff
@@ -1601,7 +1735,7 @@ Each of the steps is denoted by its numeral in parentheses, e.g. (3).
 ```{index} Mauna Loa
 ```
 
-**Mauna Loa Atmospheric CO$_{\text{2}}$ Measurements:** (1) Many 
+**Mauna Loa Atmospheric CO$_{\text{2}}$ Measurements:** (1) Many
 current forms of energy generation and conversion&mdash;from automotive
 engines to natural gas power plants&mdash;rely on burning fossil fuels and produce
 greenhouse gases, typically primarily carbon dioxide (CO$_{\text{2}}$), as a
```
trevorcampbell, line 1772: Another `{-}` in a subsection title below here ("Choose the right output...")

source/viz.md
```diff
@@ -1703,22 +1837,22 @@ bad, while raster images eventually start to look "pixelated."
...
-`chart.save` works by taking the path to the directory where you would like to save the file 
+to save a file containing this image.
+`chart.save` works by taking the path to the directory where you would like to save the file
 (e.g., `img/filename.png` to save a file named `filename` to the `img` directory),
 The kind of image to save is specified by the file extension.
-For example, 
+For example,
 to create a PNG image file, we specify that the file extension is `.png`.
-Below we demonstrate how to save PNG and SVG file types 
+Below we demonstrate how to save PNG and SVG file types
```
trevorcampbell: The code fails here. No enabled saver that supports outputting `png` (and I imagine `svg` will fail too, once it gets to that line). 

Let me know if we need to install another package in our docker container to get it to work.

source/viz.md
```diff
@@ -1792,8 +1926,8 @@ Zoomed in `faithful`, raster (PNG, left) and vector (SVG, right) formats.
 
 ## Exercises
 
-Practice exercises for the material covered in this chapter 
-can be found in the accompanying 
+Practice exercises for the material covered in this chapter
+can be found in the accompanying
 [worksheets repository](https://github.com/UBC-DSCI/data-science-a-first-intro-worksheets#readme)
```
trevorcampbell: This link is for the R version -- needs to be fixed to be the python repo

See #56

trevorcampbell (changes requested):
Made it through this whole chapter @lheagy . Nothing too major -- this one should be easier than Ch3.

source/viz.md
lheagy: This statement is pretty different than the R version, so I am not entirely sure what you have in mind here. I instead moved this statement to the preabmle. See if you think that works

source/viz.md
lheagy: I moved this to the section about faithful so mark point is relevant

source/viz.md
lheagy: I will just reword it as an option

source/viz.md
lheagy: this is perhaps a question for @joelostblom?? Altair is fairly new to me too. Any suggestions appreciated!

trevorcampbell: Looks like it is standard:

https://altair-viz.github.io/user_guide/encoding.html#encoding-data-types

in that case, @lheagy it would probably be a good idea to provide a table summarizing the different types and codes for them (which appears on that webpage I linked)

trevorcampbell:
@lheagy one more that I bumped into in Ch5: apparently, Altair's default colour palette is colourblind-friendly. Did we mention this when discussing colourblind stuff in ch4?

trevorcampbell:
This PR:
- closes #40 
- closes #25

source/viz.md
lheagy: +1 updated in line 645

source/viz.md
lheagy: resolved with 1104

source/viz.md
```diff
@@ -1086,15 +1147,15 @@ The default label is the name of the column being mapped to `color`. Here that
...
-islands_plot_sorted = alt.Chart(islands_top12).mark_bar(color='black').encode(
-    x = alt.X("size",title = "Size (1000 square mi)"),
-    y = alt.Y("landmass", title = "Landmass", sort='x'),
-    color = alt.Color("landmass_type", title = "Type")).configure_axis(
-    titleFontSize=12)
+islands_plot_sorted = alt.Chart(islands_top12).mark_bar().encode(
+    x=alt.X("size",title="Size (1000 square mi)"),
+    y=alt.Y("landmass", title="Landmass", sort='x'),
+    color=alt.Color("landmass_type", title="Type")
+).configure_axis(titleFontSize=12)
```
lheagy, line 1158: I am inclined to remove the use of `configure_axis` here. The default font size is reasonable

source/viz.md
lheagy: resolved with 1104

source/viz.md
lheagy: @joelostblom: could you give us a hand here, this is not one I am overly familiar with

source/viz.md
lheagy: updated -- line 262

source/viz.md
```diff
@@ -887,49 +965,59 @@ glue('can_lang_plot_legend', can_lang_plot_legend, display=False)
...
-this makes the scatter point shapes different for each category. This kind of 
+this makes the scatter point shapes different for each category. This kind of
 visual redundancy&mdash;i.e., conveying the same information with both scatter point color and shape&mdash;can
 further improve the clarity and accessibility of your visualization.
-
-> Note: We cannot use different shapes with `mark_circle`, it can only be used with `mark_point`
+You can use
+this [color blindness simulator](https://www.color-blindness.com/coblis-color-blindness-simulator/) to check
+if your visualizations are color-blind friendly.
+The default color palattes in `altair` are color-blind friendly (one more reason to stick with the defaults!).
```
lheagy, line 1015: added content on altair defaults being color-blind friendly

source/viz.md
```diff
@@ -1456,117 +1614,94 @@ On the other hand, the `maxbins=200` and `maxbins=5` are too small and too big,
...
 
 :::{glue:figure} final_plot_max_bins
-:figwidth: 700px 
+:figwidth: 700px
 :name: final_plot_max_bins
 
 Effect of varying number of max bins on histograms.
 :::
 
-#### Adding layers to a `altair` plot object {-}
```
lheagy, line 1500: Moved earlier on: 1306

Issue #65, opened 2022-12-19
"method", "attribute", "function", etc


trevorcampbell:
At some point (once the major rounds of polish are done), we should make a decision on:
- what convention we want for using the terms method, function, attribute, object, member, etc
- when we introduce that explicitly (do we just gloss over it at the beginning without explaining and then introduce later explicitly? do we introduce explicitly when we first mention? etc)

This is definitely a low-priority item for after the first run in the Jan 2023 semester.

joelostblom:
I was worried using the term `method` in the textbook because I think we never properly introduce it, but going through the worksheets I noticed that there is a mention in the into worksheet:

> The upper method we used here is different from the functions we used previously (e.g. print). This method is called using the dot notation (string.upper()), because this method only works for a particular kind of object that they were designed to work for. Here the upper method was written to only work with string objects. print function, however, was written to work with many kinds of objects, therefore, we don't use the dot notation.

trevorcampbell:
See also #243

PR #78, opened 2022-12-31
Chapter 5 production polish

### LINKED ISSUE #3: classification1
PR #7

### LINKED ISSUE #41: Suggestions for Ch5 - Classification 1

- [x] Should we cover pandas categoricals? Not necessary for classification but it is the equivalent to factors in R
- [x] Instead of `groupby.count` with lots of intermediate objects, we should use `.value_counts` with and without `normalize=True`.
- [x] Altair's default color scheme is colorblind friendly so no need to change here as we have to when using ggplot.
- [x] Students have not learned how to use apply + lambda... the easiest here would be to map a dictionary to the series, but I don't think we teach that either. We could change the labels as in R, but see the point below
- [x] R&Py I think it is fragile to overwrite the data values in the legend and would not recommend anyone to do this. What if the order of the data changes, either because new data is added or we do some filtering? Then we might be mistakenly labeling `B` as `Malign` and there would be no easy way to find out. I think it is much better to change the values of the individual data points using dictionary like assignment that is robust to data reordering and therefore more reproducible.
- [x] R&Py "Based on our visualization, it seems like the prediction of an unobserved label might be possible." weird sentence
- [ ] Fig 62-71 missing (and many more, need to go through the entire page and add figs)
- [x] Use method chaining  to comp distance. Proper indentation. We haven't taught assign with lambda df so need to be first. `np` not defined. This is both for 2 and more predictor variables code chunks. HTML also not defined. Also use `nlargest` instead of sorting first.
- [x] R&Py 5.5.3 I don't think it is correct to describe this as 4 steps. Step 2 and 3 should be combined into "Find the K closest points", since we actually don't care about sorting the distances that are really far away.
- [x] Unintuitive usage of loc with : to select columns for cancer_train
- [x] We don't need to mention weight = 'uniform' since that is the default? And not get into other types of weights here either
- [x] Say something more about what happens after fitting so that students realize that knn_spec changes
- [x] `unscaled_cancer` can be assigned in one step.
- [x] `make_column_transformer` not defined (imports seem to generally not be working as expected here)
- [x] "Can be combined into `fit_transform` requires more explanation (as does `transform` itself)
- [x] Explain why we are calling `pd.DataFrame` (some things in this cell seems unnecessary)
- [ ] TODO Compare with how I introduced these concepts in BAIT 509.
- [x] R `bind_rows` never explained, setting labels manually again
- [x] `pd.concat` never explained, neither is `query` nor `lambda` in `apply`.
- [x] `value_counts` instead of `groupby.count` (in text and code)
- [x] Incorrect usage of `loc[:, ...]` again in make_pipline
- [x] TODO go through prediction plot code in details (if we just include the screenshots from R for the other figures, do we really need the code for how to create one of these charts in this figure?)
- [x] Split multiple arguments over multiple rows so that they are easier to read (this goes for all chapters), e.g. 
    ```
    knn_fit = make_pipeline(preprocessor, knn_spec).fit(
        X=unscaled_cancer.loc[:, ["Area", "Smoothness"]], y=unscaled_cancer["Class"]
    )
    ```
    should be:
    ```
    knn_fit = make_pipeline(preprocessor, knn_spec).fit(
        X=unscaled_cancer.loc[:, ["Area", "Smoothness"]],
        y=unscaled_cancer["Class"]
    )
    ```
    and 
    ```
    pd.DataFrame({"Area": [500, 1500], "Smoothness": [0.075, 0.1]})
    ```
    should be:
    ```
    pd.DataFrame({
        "Area": [500, 1500],
        "Smoothness": [0.075, 0.1]
    })
    ```

### LINKED ISSUE #70: Suggestions for Ch6 - Classification 2
- [x] np.random.seed is not recommended anymore
    - [x] Instead we shuold create a random state object and use it methods
    - [x] However, I don't think we will ever need to use np.random.seed. With both sklearn methods and pandas `sample` we will always control randomness via the `random_state` parameter.
    - [x] I think it is more confusing than helpful to also talk about seeds here, and we can just show the effect of a different `random_state` when doing the `train_test` split instead.
- [x] "only function that uses randomness in **Python**"
- [x] We haven't taught how to use a lambda function like this. We could since it is useful and very flexible, or if we think it is too complex, we could teach how to perform the same operation either via `cancer["Class"].map({'M': 'Malignant', 'B': 'Benign'})` (which would be convenient but a new thing) or simply filtering: `cancer.loc[cancer['Class'] == 'M', 'Class'] = 'Malignant'` + the benign version (which is less elegant but builds on what they have learned already)
- [x] We are setting the seed in this cell but there is no randomness to control...
- [x] We don't need to set the color range
- [x] Downgrade to pandas <1.5 to avoid warning until next Altair release ` FutureWarning: iteritems is deprecated and will be removed in a future version. Use .items instead.`
- [ ] Explain why it is important to use `stratify` and `shuffle`, not just mention what they do (include a short example for each in a sentence or so)
- [x] This sentence is meant to include the actual number of observations "We can see from .info() in the code above that the training set contains observations, while the test set contains observations"
- [x] The `.info` method always prints columns vertically, not just when there is a large number of them.
- [x] We should not use `groupby` + `count` here. That creates a count for each column, so it doesn't generalize well. Instead we should use `.groupby` + `size` OR just use `value_counts(normalize=True)` which means we don't have to divide by the length after (btw we should get dataframe length via `df.shape[0]` not using `len` (applies to the entire book)). I don't think there is a `normalize` option to `size`, but can't double check because I'm in the air
    - [x] We also shouldn't create a new dataframe unless we are using it again later.
- [x] This sentence is meant to include actual numbers: "we see about % of the training data are benign and % are malignant"
- [x] Imports seems to be missing for all sklearn stuff.
- [x] Could we just call it "model" instead of "model specification" everywhere? Make it flow better so it is easier to read.
- [x] Some introduction to why we use captical `X` and lower case `y`?
    - [ ] `X` should be created as `cancer_train.drop(columns='Class')`, which is more generalizable
    - [ ] We could possibly even skip this throughout the book and just pass the dataframe?
        - [ ] If we keep it, we should call it `X_train` and do the same for the test data.
        - [ ] I think we should probably keep the assignment because we used them a few times later, like when creating `X_tune` / `y_tune` in grdisearch, which should just be the same `_train` instead
- [x] Comments about "weights uniform" and "hidden seed"?
- [x] Why do we use `concat` to add the predictions? Can't we just create a copy of the new columns the regular way in the original dataframe (if it is OK to modify at this point)? Or use `df.assign(pred=df['predictions'])` to just print it without overwriting the variable (if we have taught `assign` before here.
    - [x] We should also not overwrite variable names like this in my opinion.
    - [x] Maybe something like this:
        ```py
        cancer_test_predictions = cancer_test.copy()
        cancer_test_predictions['predicted'] = knn_fit.predict(cancer_test.drop(columns='Class'))
        ```
- [x] X_test is created in the accuracy section but was needed in the previous code cell.
- [x] I think we can do a better job explaining accuracy and start by computed it manually using `y_test` and the predictions created in `cancer_test_predictions`, so that students get a better grasp of what it is.
- [x] Missing actual value in "The output shows that the estimated accuracy of the classifier on the test data was %."
    - [x] I just realized that all of these might be fixed when the code actually runs...
- [x] Do we ever even need to show `confusion_matrix` when we are saying that `CnfusionMatrixDisiplay` is the recommend way?
    - [x] This is one of the many places where it feels unnecessarily noise to assign to a variable instead of just displaying the plot.
- [x] We need to explain confusion matrices and what is in each square, there are resources in BAIT 509 and DSCI 573 for this.
- [x] We talk about "majority classifier" but we never show how to create one. This is really easy to setup in sklearn and it is a useful strategy for students to learn so that they can compare against a baseline. I suggest we show them how to setup a dummy classifier (and a dummy regressor in later chapters). We might also consider renaming "majority classifier" to "dummy classifier" in the text so not be ambiguous
- [x] The code in the cross-validation section repeats some of the things that were done earlier in the chapter like the train_test_split
- [x] Cross-validation fig72 legend overlaps equation
- [x] Quoting the docs for the `cv` param of `cross_validate` does not seem that helpful?
- [x] The first time we use cross_validate, we do many things at once (cros_validate, reutrn_train, convert to pandas dataframe, set cv to 5 although its the default). It could be better to do these one at a time.
- [x] "We can then aggregate the mean and standard **deviation**". Again Search and replace in the entire chapter.
    - [x] We don't need `func=` in `agg`
- [x] "provides **two** build-in methods"
- [x] I think we should introduce grid search and randomized search one at a time to make it easier to understand and also include more explanation as well as an illustration. We can take this from BAIT 509 https://bait509-ubc.github.io/BAIT509/lectures/lecture6.html#automated-hyperparameter-optimization
- [x] Do we cover what `range` does somewhere or do we just start using it?
- [x] "the **cross-validation** accuracy actually starts to decrease!
- [x] `param_grid_lots` is not a great name, something like `large_param_grid` sounds better. same with the plots name.
- [x] Do we explain `n_jobs=-1`?
- [x] Rewrite this "We can evaluate a classifier by splitting the data randomly into a training and test data set, using the training set to build the classifier, and using the test set to estimate its accuracy. Finally, we can tune the classifier (e.g., select the number of neighbors in K-NN) by maximizing estimated accuracy via cross-validation." **to this** "We can tune and evaluate a classifier by splitting the data randomly into a training and test data set. The training set is used to build the classifier and we can tune the classifier (e.g., select the number of neighbors in K-NN) by maximizing estimated accuracy via cross-validation. After we have tuned the model we can use the test set to estimate its accuracy."
- [x] Swap step 2 and 3, since 2 and 4 belong together and 3 is independent of the two.
- [x] Change step 4 to read: "Setup GridSearchCV (or RandomizedSearchCV) to estimate the classifier accuracy for a range of parameter values (values of "K" in our case)."
- [x] Change step 5 to read "Execute the grid search by passing the training data to the `fit()` function of GridSearchCV instance.
- [x] "yields a high **cross-validation accuracy** ~~estimate~~ that"
- [x] Relating to step 6 and 7, GridSearchCV already provides us with the best model as an attribute, granted hat we still want them to look at the plot, is there value in teaching how to extract the already fit best model from `GridSearchCV.best_estimator_` instead of refitting it (which is redundant and potentially time consuming).
- [x] We don't show how we create the irrelevant columns
- [x] "Forward selection ~~in Python~~"
    - [x] In general I think this section is not that well-written and we should go through it and improve the language as well as how the different code snippets are introduced.

trevorcampbell:
Ignore the branch name -- I originally was planning to do both ch5+6 in this PR, but figured it would be better to just split them.

- closes #41 
- closes #3


source/classification1.md
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
 ```{index} info
 ```
 
-Below we use `.info()` to preview the data frame. This method can 
-make it easier to inspect the data when we have a lot of columns, 
-as it prints the data such that the columns go down 
-the page (instead of across).
+Below we use the `info` method to preview the data frame. This method can
```
joelostblom, line 192: I *think* we have been calling it "function" everywhere so far (for simplicity even if it is incorrect). Maybe something we can do a pass on at the end and see that we are consistent. I guess we have https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/65 for this already

source/classification1.md
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
...
+```{index} replace
+```
+
 ```{code-cell} ipython3
-:tags: [remove-cell]
+cancer['Class'] = cancer['Class'].replace({
+				'M' : 'Malignant',
+				'B' : 'Benign'
+				})
+cancer['Class'] = cancer['Class'].astype('category')
```
joelostblom:
```suggestion
cancer['Class'] = cancer['Class'].replace({
    'M' : 'Malignant',
    'B' : 'Benign'
}).astype('category')
```

I like the choice of `replace` instead of `map`. In addition to the NA issue you mentioned I think the method name if more intuitive. I think we should cover the `category` data type in the wrangling chapter so that students know why it is beneficial (although I also wonder if we will actually leverage any of the benefits in dsci100 or if it is just one more thing to learn, sorry I know I suggested it initially). https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/120

source/classification1.md
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
...
+variable. Then we calculate the percentage in each group by dividing by the total
+number of observations and multiplying by 100. We have 
+{glue:}`benign_count` ({glue:}`benign_pct`\%) benign and
+{glue:}`malignant_count` ({glue:}`malignant_pct`\%) malignant 
+tumor observations.
 
 ```{code-cell} ipython3
-num_obs = len(cancer)
 explore_cancer = pd.DataFrame()
 explore_cancer['count'] = cancer.groupby('Class')['ID'].count()
```
joelostblom, line 262:
```suggestion
cancer.groupby('Class').size()
```

There is a `size` method which I think is more straightforward since we don't have to select a column. I don't see the value in creating a new data frame here and think it is easier to understand if we just do it in two different steps, as with value_counts below

source/classification1.md
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
...
-of benign and malignant tumor observations in our data set. When paired with `.groupby()`, `.count()` counts the number of observations in each `Class` group.
-Then we calculate the percentage in each group by dividing by the total number of observations. We have 357 (63\%) benign and 212 (37\%) malignant tumor observations.
+the `groupby` and `count` methods to find the number and percentage 
+of benign and malignant tumor observations in our data set. When paired with
+`groupby`, `count` counts the number of observations for each value of the `Class`
+variable. Then we calculate the percentage in each group by dividing by the total
+number of observations and multiplying by 100. We have 
+{glue:}`benign_count` ({glue:}`benign_pct`\%) benign and
+{glue:}`malignant_count` ({glue:}`malignant_pct`\%) malignant 
+tumor observations.
```
joelostblom, line 258:
```suggestion
the `groupby` and `size` methods to find the number and percentage 
of benign and malignant tumor observations in our data set. When paired with
`groupby`, `size` counts the number of observations for each value of the `Class`
variable.
```

joelostblom: I moved the rest down so that we do one step at a time

source/classification1.md
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
...
 
 ```{code-cell} ipython3
-num_obs = len(cancer)
 explore_cancer = pd.DataFrame()
 explore_cancer['count'] = cancer.groupby('Class')['ID'].count()
-explore_cancer['percentage'] = explore_cancer['count'] / num_obs * 100
+explore_cancer['percentage'] = 100 * explore_cancer['count']/len(cancer)
 explore_cancer
 ```
```
joelostblom, line 265:
````suggestion
```

Then we calculate the percentage in each group by dividing by the total
number of observations and multiplying by 100. We have 
{glue:}`benign_count` ({glue:}`benign_pct`\%) benign and
{glue:}`malignant_count` ({glue:}`malignant_pct`\%) malignant 
tumor observations.

```{code-cell} ipython3
100 * cancer.groupby('Class').size() / cancer.shape[0]
```
````

`shape` is preferred over length on dataframes since it is already computed and in the spirit of using dataframes methods instead of global python functions in general.

File: source/classification1.md [hunk 6]
```diff
@@ -206,110 +193,114 @@ total set of variables per image in this data set is:
...
+```
+
+Next, let's draw a colored scatter plot to visualize the relationship between the
+perimeter and concavity variables. Recall that `altair's` default palette
+is colorblind-friendly, so we can stick with that here.
+
+```{code-cell} ipython3
 perim_concav = (
     alt.Chart(cancer)
     .mark_point(opacity=0.6, filled=True, size=40)
```
joelostblom:
```suggestion
    .mark_circle()
```

I think we should keep the anything that is not central to the chapter as simple as possible. In this case there is not much difference by adding these three parameters

New:
image

Old:
image

source/classification1.md
```diff
@@ -377,30 +369,28 @@ $K$ for us. We will cover how to choose $K$ ourselves in the next chapter.
...
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
 perim_concav_with_new_point = (
     alt.Chart(
         perim_concav_with_new_point_df,
-        # title="Scatter plot of concavity versus perimeter with new observation represented as a red diamond.",
     )
     .mark_point(opacity=0.6, filled=True, size=40)
```
joelostblom:
```suggestion
    .mark_circle()
```

source/classification1.md
```diff
@@ -377,30 +369,28 @@ $K$ for us. We will cover how to choose $K$ ourselves in the next chapter.
...
-            "Class",
-            scale=alt.Scale(range=["#86bfef", "#efb13f", "red"]),
-            title="Diagnosis",
-        ),
+        color=alt.Color("Class", title="Diagnosis"),
         shape=alt.Shape(
             "Class", scale=alt.Scale(range=["circle", "circle", "diamond"])
         ),
-        size=alt.condition("datum.Class == 'Unknown'", alt.value(80), alt.value(30))
+        size=alt.condition("datum.Class == 'Unknown'", alt.value(80), alt.value(30)),
```
joelostblom:
```suggestion
        size=alt.condition("datum.Class == 'Unknown'", alt.value(100), alt.value(30)),
        stroke=alt.condition("datum.Class == 'Unknown'", alt.value('black'), alt.value(None)),
```
I vote for making this new point easier to see:

After this change:
image



Previously:
image

source/classification1.md
```diff
@@ -509,9 +500,10 @@ glue("2-neighbor_con", round(near_neighbor_df2.iloc[0, :]['Concavity'], 1))
 glue('fig:05-knn-4', (perim_concav_with_new_point2 + line2), display=True)
```
joelostblom, line 496:
```suggestion
perim_concav_with_new_point['data'] = perim_concav_with_new_point_df2
line['data'] = new_neighbor_df2
glue('fig:05-knn-4', (perim_concav_with_new_point + line), display=True)
```

We don't need to recreate the charts each time with the exact same spec. It is enough to just switch out the data, so we can delete all the code for `line2` and `perim_concav_with_new_point2`; the same goes below. This will make it easier to keep this good up to date since changes only happen in one location.

I think it is harmless to overwrite the data in the same variable in this case, but if we don't want that for some reason, we can use `perim_concav_with_new_point.properties(data=perim_concav_with_new_point_df2)` to create a new chart, it's just a bit more verbose.

source/classification1.md
```diff
@@ -660,13 +660,14 @@ Scatter plot of concavity versus perimeter with new observation represented as a
...
-cancer_dist = cancer_dist.sort_values(by="dist_from_new").head(5)
+cancer_dist = (cancer
+           .loc[:, ["Perimeter", "Concavity", "Class"]]
+           .assign(dist_from_new = (
+               (cancer["Perimeter"] - new_obs_Perimeter) ** 2
+             + (cancer["Concavity"] - new_obs_Concavity) ** 2
+           )**(1/2))
+           .nsmallest(5, "dist_from_new")
+      )
 cancer_dist
```
joelostblom:
```suggestion
(
    cancer
   [["Perimeter", "Concavity", "Class"]]
   .assign(dist_from_new = (
       (cancer["Perimeter"] - new_obs_Perimeter) ** 2
     + (cancer["Concavity"] - new_obs_Concavity) ** 2
   )**(1/2))
   .nsmallest(5, "dist_from_new")
)
```

Remove redundant assignment and unnecessary use of loc

source/classification1.md
```diff
@@ -771,18 +757,20 @@ three predictors.
...
+cancer_dist2 = (cancer
+           .loc[:, ["Perimeter", "Concavity", "Symmetry", "Class"]]
+           .assign(dist_from_new = (
+               (cancer["Perimeter"] - new_obs_Perimeter) ** 2
+             + (cancer["Concavity"] - new_obs_Concavity) ** 2
+             + (cancer["Symmetry"] - new_obs_Symmetry) ** 2
+           )**(1/2))
+           .nsmallest(5, "dist_from_new")
+      )
 cancer_dist2
```
joelostblom:
```suggestion
(
    cancer
    [["Perimeter", "Concavity", "Symmetry", "Class"]]
    .assign(dist_from_new = (
        (cancer["Perimeter"] - new_obs_Perimeter) ** 2
        + (cancer["Concavity"] - new_obs_Concavity) ** 2
        + (cancer["Symmetry"] - new_obs_Symmetry) ** 2
    )**(1/2))
    .nsmallest(5, "dist_from_new")
)
```

source/classification1.md
```diff
@@ -1101,19 +1039,26 @@ is said to be *standardized*, and all variables in a data set will have a mean o
...
-unscaled_cancer = unscaled_cancer[['Class', 'Area', 'Smoothness']]
+unscaled_cancer = (
+            pd.read_csv("data/unscaled_wdbc.csv")
+            .loc[:, ['Class', 'Area', 'Smoothness']]
+            .replace({
+               'M' : 'Malignant',
+               'B' : 'Benign'
+            })
+    )
+unscaled_cancer['Class'] = unscaled_cancer['Class'].astype('category')
```
joelostblom:
```suggestion
unscaled_cancer = pd.read_csv("data/unscaled_wdbc.csv")[['Class', 'Area', 'Smoothness']]
unscaled_cancer['Class'] = unscaled_cancer['Class'].replace({
   'M' : 'Malignant',
   'B' : 'Benign'
}).astype('category')
unscaled_cancer
```

I think we should be careful showing replace on the entire dataframe without talking about it explicitly. It is easy to make a mistake and replace values in columns that were not intended to be replaced, so it's safer to limit it to one column.

source/classification1.md
```diff
@@ -1159,68 +1100,74 @@ So far, we have built a preprocessor so that each of the predictors have a mean
...
-# centering (`step_center`) \index{recipe!step\_center} steps for 
-# all of the predictors so that they each have a mean of 0 and standard deviation of 1.
-# Note that `tidyverse` actually provides `step_normalize`, which does both centering and scaling in
-# a single recipe step; in this book we will keep `step_scale` and `step_center` separate
-# to emphasize conceptually that there are two steps happening. 
-# The `prep` function finalizes the recipe by using the data (here, `unscaled_cancer`)  \index{tidymodels!prep}\index{prep|see{tidymodels}}
-# to compute anything necessary to run the recipe (in this case, the column means and standard
-# deviations):
+preprocessor = make_column_transformer(
+           (StandardScaler(), make_column_selector(dtype_include=np.number)),
```
joelostblom: Note that we could simply use `'number'` here instead of importing numpy. The drawback is that there is no error directly in this step if there is a typo so it could be confusing. The upside is that we could specify `'category'` the same way to select all pandas categories. See the notes section here https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.select_dtypes.html#pandas.DataFrame.select_dtypes

source/classification1.md
```diff
@@ -1159,68 +1100,74 @@ So far, we have built a preprocessor so that each of the predictors have a mean
...
+```
+It looks like our `Smoothness` and `Area` variables have been standardized. Woohoo!
+But there are two important things to notice about the new `scaled_cancer` data frame. First, it only keeps
+the columns from the input to `transform` (here, `unscaled_cancer`) that had a preprocessing step applied
+to them. The default behavior of the `ColumnTransformer` that we build using `make_column_transformer` 
+is to *drop* the remaining columns. This default behavior works well with the rest of `sklearn` (as we will see below
+in the {ref}`08:puttingittogetherworkflow` section), but for visualizing the result of preprocessing it can be useful to keep the other columns
+in our original data frame, such as the `Class` variable here.
+To keep other columns, we need to set the `remainder` argument to `'passthrough'` in the `make_column_transformer` function.
+ Furthermore, you can see that the new column names---{glue:}`scaled_cancer.columns[0]`
```
joelostblom: The glued variables are now showing up for me

image

source/classification1.md
```diff
@@ -1523,39 +1430,25 @@ what the data would look like if the cancer was rare. We will do this by
...
 rare_cancer = pd.concat(
-    (cancer.query("Class == 'B'"), cancer.query("Class == 'M'").head(3))
-)
-colors = ["#86bfef", "#efb13f"]
-rare_cancer["Class"] = rare_cancer["Class"].apply(
-    lambda x: "Malignant" if (x == "M") else "Benign"
-)
+    (cancer[cancer["Class"] == 'Benign'],
+     cancer[cancer["Class"] == 'Malignant'].head(3)
+    ))
```
joelostblom:
```suggestion
    cancer[cancer["Class"] == 'Benign'],
    cancer[cancer["Class"] == 'Malignant'].head(3)
))
```

Nice this new approach is much more intuitive than the one with `apply`! I moved this parent up one line

source/classification1.md
```diff
@@ -1523,39 +1430,25 @@ what the data would look like if the cancer was rare. We will do this by
...
-```
+We use the [`concat`](https://pandas.pydata.org/docs/reference/api/pandas.concat.html) 
+function from `pandas` to glue the two resulting filtered
+data frames back together by passing them together in a sequence.
+The new imbalanced data is shown in {numref}`fig:05-unbalanced`, 
+and we print the counts of the classes using the `value_counts` function.
 
 ```{code-cell} ipython3
-cancer = pd.read_csv("data/wdbc.csv")
 rare_cancer = pd.concat(
```
joelostblom:
```suggestion
rare_cancer = pd.concat((
```

source/classification1.md
```diff
@@ -1523,39 +1430,25 @@ what the data would look like if the cancer was rare. We will do this by
...
+    (cancer[cancer["Class"] == 'Benign'],
+     cancer[cancer["Class"] == 'Malignant'].head(3)
+    ))
+
 rare_plot = (
-    alt.Chart(
-        rare_cancer
-    )
+    alt.Chart(rare_cancer)
     .mark_point(opacity=0.6, filled=True, size=40)
```
joelostblom:
```suggestion
    .mark_circle()
```

source/classification1.md
```diff
@@ -1705,10 +1604,10 @@ prediction_plot = (
                 domain=(rare_cancer["Concavity"].min(), rare_cancer["Concavity"].max())
             ),
         ),
-        color=alt.Color("Class", scale=alt.Scale(range=colors), title="Diagnosis"),
+        color=alt.Color("Class", title="Diagnosis"),
     )
 )
-rare_plot + prediction_plot
+#rare_plot + prediction_plot
 glue("fig:05-upsample-2", (rare_plot + prediction_plot))
```
joelostblom, line 1606: Hmm, for some reason I see the wrong chart in the book I build locally here:

image

Ah, the reason is that the glue variable is overwritten in the last chart. see my edit just before the exercises for fixing this

source/classification1.md
```diff
@@ -1988,13 +1855,17 @@ prediction_plot = (
...
+        color=alt.Color("Class", title="Diagnosis"),
     )
 )
-
 unscaled_plot + prediction_plot
 ```
 
+```{code-cell} ipython3
+:tags: [remove-input]
+glue("fig:05-upsample-2", (unscaled_plot + prediction_plot))
```
joelostblom:
```suggestion
glue("fig:05-upsample-3", (unscaled_plot + prediction_plot))
```

This overwrites a figure with the same name above. Maybe the figure name below should be changed too and use this one? Or where is this figure meant to be used?

source/classification1.md
```diff
@@ -1727,27 +1626,16 @@ Despite the simplicity of the problem, solving it in a statistically sound manne
...
+voting power in the $K$-nearest neighbor algorithm. In order to do this, we will 
+first separate the classes out into their own data frames by filtering.
+Then, we will
+use the [`resample`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.resample.html) function 
+from the `sklearn` package to increase the number of `Malignant` observations to be the same as the number 
+of `Benign` observations. We set the `n_samples` argument to be the number of `Malignant` observations we want. 
+We also set the `random_state` to be some integer
+so that our results are reproducible; if we do not set this argument, we will get a different upsampling each time
+we run the code. Finally, we use the `value_counts` method
+ to see that our classes are now balanced.
```
joelostblom: Here we introduce random state, then in the next chapter we talk about seeds instead.

source/classification1.md
```diff
@@ -1159,68 +1100,74 @@ So far, we have built a preprocessor so that each of the predictors have a mean
...
- For example, we may want to specify new data that were not part of the training set. 
+> **Note:** Only specify the `remainder` and `verbose_feature_names_out` arguments when you want to examine the result
+> of your preprocessing step. In most cases, you should leave these arguments at their default values.
+
+```{code-cell} ipython3
+preprocessor_keep_all = make_column_transformer(
+           (StandardScaler(), make_column_selector(dtype_include=np.number)),
+            remainder='passthrough',
+            verbose_feature_names_out=False
+      )
```
joelostblom:
```suggestion
preprocessor_keep_all = make_column_transformer(
    (StandardScaler(), make_column_selector(dtype_include=np.number)),
    remainder='passthrough',
    verbose_feature_names_out=False
)
```

source/classification1.md
joelostblom:
```suggestion
    (StandardScaler(), make_column_selector(dtype_include=np.number)),
```

source/classification1.md
```diff
@@ -1159,68 +1100,74 @@ So far, we have built a preprocessor so that each of the predictors have a mean
...
-# all of the predictors so that they each have a mean of 0 and standard deviation of 1.
-# Note that `tidyverse` actually provides `step_normalize`, which does both centering and scaling in
-# a single recipe step; in this book we will keep `step_scale` and `step_center` separate
-# to emphasize conceptually that there are two steps happening. 
-# The `prep` function finalizes the recipe by using the data (here, `unscaled_cancer`)  \index{tidymodels!prep}\index{prep|see{tidymodels}}
-# to compute anything necessary to run the recipe (in this case, the column means and standard
-# deviations):
+preprocessor = make_column_transformer(
+           (StandardScaler(), make_column_selector(dtype_include=np.number)),
+      )
```
joelostblom:
```suggestion
)
```

joelostblom: In case this large indentation on open params happens because of the default vim settings, which was something that was driving me nuts for a long while, you can set `let g:pyindent_open_paren = 'shiftwidth()'` as per https://stackoverflow.com/questions/23016531/how-to-make-vim-only-insert-4-spaces-one-indent-level-for-the-next-line-after/74711118#74711118

source/classification1.md
```diff
@@ -1827,101 +1715,79 @@ Upsampled data with background color indicating the decision of the classifier.
...
+# load the unscaled cancer data, make Class readable
+unscaled_cancer = (
+            pd.read_csv("data/unscaled_wdbc.csv")
+            .replace({
+               'M' : 'Malignant',
+               'B' : 'Benign'
+            })
+    )
+# make Class a categorical type
+unscaled_cancer['Class'] = unscaled_cancer['Class'].astype('category')
```
joelostblom:
```suggestion
unscaled_cancer = pd.read_csv("data/unscaled_wdbc.csv")
unscaled_cancer['Class'] = unscaled_cancer['Class'].replace({
   'M' : 'Malignant',
   'B' : 'Benign'
}).astype('category')
unscaled_cancer
```

joelostblom (changes requested):
Looking great overall, nicely done!

source/classification1.md
```diff
@@ -12,41 +12,26 @@ kernelspec:
...
 
-(classification)=
-# Classification I: training & predicting
-
 ```{code-cell} ipython3
 :tags: [remove-cell]
+import warnings
+def warn(*args, **kwargs):
+    pass
+warnings.warn = warn
```
joelostblom:
```suggestion
import warnings
warnings.filterwarnings("ignore", category=DeprecationWarning)
```

Does this work for you? I left a note in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/80 too

source/classification1.md
```diff
@@ -1827,101 +1716,79 @@ Upsampled data with background color indicating the decision of the classifier.
...
 ```{code-cell} ipython3
 new_observation = pd.DataFrame({"Area": [500, 1500], "Smoothness": [0.075, 0.1]})
 prediction = knn_fit.predict(new_observation)
 prediction
 ```
 
-The classifier predicts that the first observation is benign ("B"), while the second is
-malignant ("M"). {numref}`fig:05-workflow-plot-show` visualizes the predictions that this 
+The classifier predicts that the first observation is benign, while the second is
+malignant. {numref}`fig:05-workflow-plot-show` visualizes the predictions that this
```
lheagy: below uses `numpy.meshgrid`. I would be inclined to replace with `np.meshgrid` (or say "the `meshgrid` function from `numpy`)

lheagy (commented):
@joelostblom did a thorough review -- I just added a minor comment (and a commit that enables scroll for the cancer data table because it is quite large). It otherwise looks good to me!

source/classification1.md
trevorcampbell: Yep, I'll punt this to #65

source/classification1.md
trevorcampbell: Let's stick with `category` -- that data type itself isn't such a huge deal, but the important lesson for students to learn is that they should care about data types generally.

I like your suggestion on the code block, will make this change everywhere it's relevant

source/classification1.md
trevorcampbell: good point. will edit that to remove mention and hide a call to `np.random.seed` internally.

source/classification1.md
trevorcampbell: I tried this originally, and it didn't work. Let me try it again now.

source/classification1.md
trevorcampbell: agree!

source/classification1.md
trevorcampbell: ah. Duh. the trick is to ignore `FutureWarning`, not `DeprecationWarning`. Will edit this and other chapters.

source/classification1.md
trevorcampbell: nah, I just caught the same issue and fixed it -- you're right, it was buggy.

there were actually some other buggy figs (color/point size was off) so I've fixed those too

source/classification1.md
trevorcampbell: yep, I fixed that just now before looking at your review :)

source/classification1.md
trevorcampbell: I like your suggestion over the numpy version. Will change.

source/classification1.md
trevorcampbell: fixed

source/classification1.md
trevorcampbell: Opened an issue about converting all to shape throughout

I will open another issue about size -- we may need to edit earlier chapters to do that

source/classification1.md
trevorcampbell: Something was off about this... it didn't work for me. We can do this later, it's good enough for now as is

PR #86, opened 2023-01-02
Chapter 6 production polish

### LINKED ISSUE #9: classification2
Branch: [classification2](https://github.com/UBC-DSCI/introduction-to-datascience-python/tree/classification2)
PR opened: PR #10 , the [deployed Chapter](https://gloriawyy.github.io/classification2/classification2.html)

### LINKED ISSUE #70: Suggestions for Ch6 - Classification 2

trevorcampbell:
- closes #70 
- closes #9


trevorcampbell:
@joelostblom there is one figure at the end (Figure 6.7) that is not generating properly --- see the R version to see what the plot should look like). Your help there would be much appreciated.

Aside from that @joelostblom and @lheagy this is ready for review.

Please review the Ch5 PR first #78 since this builds on that one.

source/classification2.md
```diff
@@ -56,7 +42,7 @@ By the end of the chapter, readers will be able to do the following:
 - Describe what training, validation, and test data sets are and how they are used in classification.
 - Split data into training, validation, and test data sets.
 - Describe what a random seed is and its importance in reproducible data analysis.
-- Set the random seed in Python using the `numpy.random.seed` function or `random_state` argument in some of the `scikit-learn` functions.
```
joelostblom, line 59: See the note I added in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/114 for some more thoughts on this.

source/classification2.md
```diff
@@ -317,23 +320,24 @@ np.random.seed(1)
...
-cancer["Class"] = cancer["Class"].apply(
-    lambda x: "Malignant" if (x == "M") else "Benign"
-)
+# re-label Class 'M' as 'Malignant', and Class 'B' as 'Benign',
+# and change the Class variable to have a category type
+cancer['Class'] = cancer['Class'].replace({
+				'M' : 'Malignant',
+				'B' : 'Benign'
+				})
+cancer['Class'] = cancer['Class'].astype('category')
```
joelostblom:
```suggestion
cancer['Class'] = cancer['Class'].replace({
    'M' : 'Malignant',
    'B' : 'Benign'
}).astype('category')
```

source/classification2.md
```diff
@@ -180,115 +164,134 @@ The trick is that in Python&mdash;and other programming languages&mdash;randomne
...
 
 np.random.seed(1)
-random_numbers = np.random.choice(range(10), size=10, replace=True)
-random_numbers
-```
 
-You can see that `random_numbers` is a list of 10 numbers
+nums_0_to_9 = pd.Series([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+
+random_numbers1 = nums_0_to_9.sample(n = 10).values
```
joelostblom: `values` will be deprecated in favor of `to_numpy()` so it would be preferred if we use that I think

source/classification2.md
```diff
@@ -317,23 +320,24 @@ np.random.seed(1)
...
+cancer['Class'] = cancer['Class'].astype('category')
 
 # create scatter plot of tumor cell concavity versus smoothness,
 # labeling the points be diagnosis class
-## create a list of colors that will be used to customize the color of points
-colors = ["#86bfef", "#efb13f"]
 
 perim_concav = (
     alt.Chart(cancer)
     .mark_point(opacity=0.6, filled=True, size=40)
```
joelostblom:
```suggestion
    .mark_circle()
```

source/classification2.md
```diff
@@ -434,42 +419,30 @@ glue("cancer_test_nrow", len(cancer_test))
...
+in `cancer_train`. We see about {glue:}`cancer_train_b_prop`% of the training
 data are benign and {glue:}`cancer_train_m_prop`% 
 are malignant, indicating that our class proportions were roughly preserved when we split the data.
 
 ```{code-cell} ipython3
-cancer_proportions = pd.DataFrame()
-cancer_proportions['n'] = cancer_train.groupby('Class')['ID'].count()
-cancer_proportions['percent'] = 100 * cancer_proportions['n'] / len(cancer_train)
-cancer_proportions
+cancer_train["Class"].value_counts(normalize = True)
```
joelostblom:
```suggestion
cancer_train["Class"].value_counts(normalize=True)
```

source/classification2.md
```diff
@@ -546,80 +506,50 @@ knn_fit
...
-# )
-
-# # move the `predicted` column to the first column for easy visualization
-# col_order = cancer_test_predictions.columns.tolist()
-# col_order = col_order[-1:] + col_order[:-1]
-# cancer_test_predictions[col_order]
+cancer_test_predictions = cancer_test.assign(
+                  predicted = knn_fit.predict(
+                         cancer_test.loc[:, ["Smoothness", "Concavity"]]
+                  ))
```
joelostblom:
```suggestion
cancer_test_predictions = cancer_test.assign(
    predicted = knn_fit.predict(cancer_test.loc[:, ["Smoothness", "Concavity"]])
)
```

source/classification2.md
```diff
@@ -546,80 +506,50 @@ knn_fit
...
+Finally, we can assess our classifier's accuracy. We could compute the accuracy manually
+by using our earlier formula: the number of correct predictions divided by the total
+number of predictions. First we filter the rows to find the number of correct predictions,
+and then divide the number of rows with correct predictions by the total number of rows
+using the `len` function.
 ```{code-cell} ipython3
-:tags: [remove-cell]
+correct_preds = cancer_test_predictions[
+    cancer_test_predictions['Class'] == cancer_test_predictions['predicted']
+    ]
```
joelostblom:
```suggestion
]
```

source/classification2.md
```diff
@@ -546,80 +506,50 @@ knn_fit
...
 
-# Finally, we can assess our classifier's accuracy. To do this we use the `metrics` function \index{tidymodels!metrics}
-# from `tidymodels` to get the statistics about the quality of our model, specifying
-# the `truth` and `estimate` arguments:
+len(correct_preds) / len(cancer_test_predictions)
```
joelostblom:
```suggestion
correct_preds.shape[0] / cancer_test_predictions.shape[0]
```

source/classification2.md
```diff
@@ -546,80 +506,50 @@ knn_fit
...
+the `score` method. To use the `score` method, we need to specify two arguments:
+predictors and true labels. We pass the same test data
+for the predictors that we originally passed into `predict` when making predictions,
+and we provide the true labels via the `cancer_test["Class"]` series.
 
+```{code-cell} ipython3
+cancer_acc_1 = knn_fit.score(
+                cancer_test.loc[:, ["Smoothness", "Concavity"]],
+                cancer_test["Class"]
+              )
```
joelostblom:
```suggestion
cancer_acc_1 = knn_fit.score(
    cancer_test.loc[:, ["Smoothness", "Concavity"]],
    cancer_test["Class"]
)
```

source/classification2.md
```diff
@@ -629,56 +559,35 @@ cancer_acc_1
...
-```{code-cell} ipython3
-from sklearn.metrics import ConfusionMatrixDisplay
-
-confusion_display = ConfusionMatrixDisplay(
-    confusion_matrix=confusion, display_labels=knn_fit.classes_
-)
-confusion_display.plot();
+pd.crosstab(cancer_test_predictions["Class"],
+            cancer_test_predictions["predicted"]
+           )
```
joelostblom:
```suggestion
pd.crosstab(
    cancer_test_predictions["Class"],
    cancer_test_predictions["predicted"]
)
```

source/classification2.md
```diff
@@ -732,7 +641,7 @@ As an example, in the breast cancer data, recall the proportions of benign and m
 observations in the training data are as follows:
 
 ```{code-cell} ipython3
-cancer_proportions
+cancer_train["Class"].value_counts(normalize = True)
```
joelostblom:
```suggestion
cancer_train["Class"].value_counts(normalize=True)
```

source/classification2.md
```diff
@@ -818,92 +728,67 @@ models, and evaluate their accuracy. We will start with just a single
...
-X_valid = cancer_validation.loc[:, ["Smoothness", "Concavity"]]
-y_valid = cancer_validation["Class"]
-acc = knn_fit.score(X_valid, y_valid)
+knn_fit = make_pipeline(cancer_preprocessor, knn).fit(X, y)
 
+# compute the score on validation data
+acc = knn_fit.score(
+        cancer_validation.loc[:, ["Smoothness", "Concavity"]],
+        cancer_validation["Class"]
+       )
```
joelostblom:
```suggestion
acc = knn_fit.score(
    cancer_validation.loc[:, ["Smoothness", "Concavity"]],
    cancer_validation["Class"]
)
```

source/classification2.md
```diff
@@ -1385,10 +1234,9 @@ completely different.  In general, if the model *is influenced too much* by the
 training data, it is said to **overfit** the data.
 
 ```{code-cell} ipython3
-:tags: [remove-cell]
 
-# create the scatter plot
-colors = ["#86bfef", "#efb13f"]
+alt.data_transformers.disable_max_rows()
+alt.renderers.enable("mimetype")
```
joelostblom:
```suggestion
```

This uses jupyterlab to insert a png as a backup image for charts and a vega-lite instead of html representation of the interactive version of the chart. Since jupyter-book does not call jupyterlab in anyway, the image is not create and the browser can't handle the vega-lite spec, but needs the html version so we'll leave it with the original renderer for now.

joelostblom (changes requested):
Looking good overall! My biggest comment is the one I made in the issues regarding randomness.

source/classification2.md
trevorcampbell: I'm going to stick with the current `np.random.seed` approach as discussed in #114

source/classification2.md
trevorcampbell: good call!

PR #97, opened 2023-01-05
Ch3 suggestions

### LINKED ISSUE #95: Ch3: review usage of `loc[]`
A @joelostblom comment from #76 

To me it seems like everything that is done is this section could be done with just [] and there is no need to involve loc here. We can save loc for only those situation when we are both filtering rows and selecting columns, similar to what we did in the intro chapter.

joelostblom:
This is my continuation of the review of ch3 I started yesterday. I think we could simplify this chapter quite a lot via #100 (but that's for later maybe).

close #95

### DIFF
--- a/source/wrangling.md
+++ b/source/wrangling.md
@@ -1013,60 +1013,6 @@ The query (criteria we are using to select values) is input as a string. The `qu
 is less often used than the earlier approaches we introduced, but it can come in handy
 to make long chains of filtering operations a bit easier to read.
 
-## Using `[]` to select ranges of columns
-
-Suppose we wanted to select only the columns `language`, `region`,
-`most_at_home` and `most_at_work` from the `tidy_lang` data set. Using what we
-learned in the chapter on {ref}`intro`, we would pass all of these column names into the square brackets.
-
-```{code-cell} ipython3
-:tags: ["output_scroll"]
-tidy_lang[["language", "region", "most_at_home", "most_at_work"]]
-```
-
-Note that we could obtain the same result by stating that we would like all of the columns
-from `language` through `most_at_work`. Instead of passing a list of all of the column
-names that we want, we can ask for the range of columns `"language":"most_at_work"`, which
-you can read as "The columns from `language` to `most_at_work`".
-
-```{code-cell} ipython3
-:tags: ["output_scroll"]
-tidy_lang["language":"most_at_work"]
-```
-
-Similarly, you can ask for all of the columns including and after `language` by doing the following
-
-```{code-cell} ipython3
-:tags: ["output_scroll"]
-tidy_lang["language":]
-```
-
-By not putting anything after the `:`, Python reads this as "from `language` until the last column".
-Although the notation for selecting a range using `:` is convenient because less code is required,
-it must be used carefully. If you were to re-order columns or add a column to the data frame, the
-output would change. Using a list is more explicit and less prone to potential confusion.
-
-Suppose instead we wanted to extract columns that followed a particular pattern
-rather than just selecting a range. For example, let's say we wanted only to select the
-columns `most_at_home` and `most_at_work`. There are other functions that allow
-us to select variables based on their names. In particular, we can use the `.str.startswith` method
-to choose only the columns that start with the word "most":
-
-```{code-cell} ipython3
-tidy_lang[tidy_lang.columns.str.startswith('most')]
-```
-
-```{index} pandas.Series; str.contains
-```
-
-We could also have chosen the columns containing an underscore `_` by using the
-`.str.contains("_")`, since we notice
-the columns we want contain underscores and the others don't.
-
-```{code-cell} ipython3
-tidy_lang[tidy_lang.columns.str.contains('_')]
-```
-
 (loc-iloc)=
 ## Using `loc[]` to filter rows and select columns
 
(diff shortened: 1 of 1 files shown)


source/wrangling.md
```diff
@@ -1076,46 +1064,73 @@ We could also have chosen the columns containing an underscore `_` by using the
...
+we pass `1:` after the comma
 indicating we want columns after and including index 1 (*i.e.* `language`).
 
 ```{code-cell} ipython3
-column_range = tidy_lang.iloc[:, 1:]
-column_range
+tidy_lang.iloc[:, 1:]
+```
+
+We can also use `iloc[]` to select ranges of rows, using a similar syntax.
```
joelostblom: I added this to show that rows can be subset too

source/wrangling.md
```diff
@@ -1251,52 +1266,44 @@ summary statistics that you can compute with `pandas`.
 What if you want to calculate summary statistics on an entire data frame? Well,
 it turns out that the functions in {numref}`tab:basic-summary-statistics`
 can be applied to a whole data frame!
-For example, we can ask for the number of rows that each column has using `count`.
-```{code-cell} ipython3
-region_lang.count()
```
joelostblom, line 1256: I don't think `count` is a good example since data frame must have the same number of rows in each column, and we should get that infor from shape or info instead. I think this flows better from the previous section too

source/wrangling.md
```diff
@@ -1251,52 +1266,44 @@ summary statistics that you can compute with `pandas`.
...
-some of the columns in `region_lang` contain string data with words like `"Vancouver"`
-and `"Halifax"`---for these columns there is no way for `pandas` to compute the mean.
-So we provide the keyword `numeric_only=True` so that it only computes the mean of columns with numeric values. This
-is also needed if you want the `sum` or `std`.
-```{code-cell} ipython3
-region_lang.mean(numeric_only=True)
-```
-If we ask for the `min` or the `max`, `pandas` will give you the smallest or largest number
-for columns with numeric values. For columns with text, it will return the
-least repeated value for `min` and the most repeated value for `max`. Again,
```
joelostblom, line 1268: I believe that this is incorrect; it is alphabetical.

source/wrangling.md
```diff
@@ -1251,52 +1266,44 @@ summary statistics that you can compute with `pandas`.
...
+so in this case we *must* provide the keyword `numeric_only=True`
+so that the mean is only computed on columns with numeric values.
 
-```{figure} img/summarize/summarize.003.jpeg
-:name: fig:summarize-across
-:figclass: figure
-
-`loc[]` or `apply` is useful for efficiently calculating summary statistics on
-many columns at once. The darker, top row of each table represents the column
-headers.
```
joelostblom, line 1288: I don't understand what this is means to say. `loc` is already explained in the text and `apply` is the not efficient, it is more inefficient than what we use here. The figure also does not seem to add anything, I don't understand what it is showing myself.

source/wrangling.md
```diff
@@ -1334,56 +1341,71 @@ The `groupby` function takes at least one argument&mdash;the columns to use in t
 grouping. Here we use only one column for grouping (`region`).
 
 ```{code-cell} ipython3
-region_lang.groupby("region")["most_at_home"].agg(["min", "max"])
```
joelostblom, line 1337: I think this order of explanation is more helpful, introducing one step at a time

source/wrangling.md
```diff
@@ -1334,56 +1341,71 @@ The `groupby` function takes at least one argument&mdash;the columns to use in t
...
 
+The resulting dataframe has `region` as an index name.
+This is similar to what happened when we reshaped dataframes in the previous chapter,
+and just as we did then,
+you can use `reset_index` to get back to a regular dataframe
+with `region` as a column name.
+
+```{code-cell} ipython3
+region_lang.groupby("region")["most_at_home"].agg(["min", "max"]).reset_index()
+```
```
joelostblom, line 1380: I think this is quite important to mention here too as we don't cover named indices or multi-indices

source/wrangling.md
```diff
@@ -1334,56 +1341,71 @@ The `groupby` function takes at least one argument&mdash;the columns to use in t
...
+and then call `[]` with a list of column names that includes `region`;
+this approach always works.
+
 ```{code-cell} ipython3
 :tags: ["output_scroll"]
-region_lang.loc[
-  :,
-  ["region", "mother_tongue", "most_at_home", "most_at_work", "lang_known"]
-].groupby("region").max()
+region_lang.groupby("region")[["most_at_home", "lang_known"]].max()
```
joelostblom: I think this is another example of why we should prefer `[]` over `loc` whenever possible.

source/wrangling.md
```diff
@@ -1014,58 +1014,71 @@ is less often used than the earlier approaches we introduced, but it can come in
...
 it must be used carefully. If you were to re-order columns or add a column to the data frame, the
 output would change. Using a list is more explicit and less prone to potential confusion.
 
 Suppose instead we wanted to extract columns that followed a particular pattern
 rather than just selecting a range. For example, let's say we wanted only to select the
 columns `most_at_home` and `most_at_work`. There are other functions that allow
 us to select variables based on their names. In particular, we can use the `.str.startswith` method
-to choose only the columns that start with the word "most":
+to choose only the columns that start with the word "most".
+Since the `str.starswith` expression returns a list of column names
```
trevorcampbell: typo "starswith"

source/wrangling.md
```diff
@@ -1014,58 +1014,71 @@ is less often used than the earlier approaches we introduced, but it can come in
...
 
 (loc-iloc)=
-## Using `loc[]` to filter rows and select columns.
+## Using `loc[]` to filter rows and select columns
+
 ```{index} pandas.DataFrame; loc[]
 ```
 
-The `[]` operation is only used when you want to filter rows or select columns;
+The `[]` operation is only used when you want to either filter rows **or** select columns;
```
trevorcampbell, line 1033: bit unclear given the below section on using `:` (where you need to use `.loc[]` even if you're only doing a range on one var)

source/wrangling.md
```diff
@@ -1014,58 +1014,71 @@ is less often used than the earlier approaches we introduced, but it can come in
...
+```{code-cell} ipython3
+:tags: ["output_scroll"]
+tidy_lang.loc[
+    tidy_lang['region'] == 'Toronto',
+    ["language", "region", "most_at_home", "most_at_work"]
+]
+```
+
+### Using `loc[]` to select ranges of columns
+
```
trevorcampbell: the connecting thread in this section doesn't seem to be the use of `loc[]`, it seems to be about selecting ranges

at a high level, many of the edits in this section make it flow poorly -- lots of jumping between `[]` and `.loc[]` when the section is supposed to be about "using `.loc[] to do X, when/why to use .loc[] instead of []"

I think we need to more carefully consider how we want to organize this stuff

source/wrangling.md
trevorcampbell: ugh... also complicated by the fact that you *can* use `[]` for `str.startswith` etc.

source/wrangling.md
trevorcampbell: seems like something we would want to do in the `.loc[]` section above too (assuming we didn't already)

and for `[]` by itself too somewhere, unless we've already done that elsewhere

source/wrangling.md
```diff
@@ -1251,52 +1261,44 @@ summary statistics that you can compute with `pandas`.
...
+you can first use `[]` to select those columns
+and then ask for the summary statistic,
+as we did for a single column previously:
 Lets say that we want to know
 the mean and standard deviation of all of the columns between `"mother_tongue"` and `"lang_known"`.
-We use `loc[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
+We use `[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
 ```{code-cell} ipython3
-region_lang.loc[:, "mother_tongue":"lang_known"].agg(["mean", "std"])
+region_lang["mother_tongue":"lang_known"].agg(["mean", "std"])
```
trevorcampbell: wait, I thought `:` wasn't allowed in `[]` as per text above

trevorcampbell (commented):
I put in these comments after just reading the diff. Usually that's sufficient, but in this case I think I need to do another pass on this with more comments after reading both the compiled original and modified chapter

source/wrangling.md
trevorcampbell: e.g. I'm OK with demonstrating various uses of `.loc[]` that flow better with the text pedagogically even if technically one could use `[]` for the same purpose.

source/wrangling.md
joelostblom: `[]` by itself comes just after `iloc` below. If I remember correctly, I opted to not include row ranges with `loc` because then we need to get into explaining named indexes. For example `.loc[:5]` is only the same as `[:5]` and `.iloc[:5]` if the index names of the first five rows are 0,1,2,3,4, since `loc` always selects by row/index name/label.

source/wrangling.md
joelostblom: Good catch, this is a typo and should be `region_lang.loc[:, "most_at_home":"most_at_work"]` as in the next section.

source/wrangling.md
joelostblom: The only jump to `[]` that I see is when we use `starswith` and that was a mistake on my part because we should be using `loc` there. I updated this section, do you think it flows better now?

source/wrangling.md
joelostblom: When we use `loc` with a range we are technically filtering rows too with the `:`, but I see what you mean regarding that the intention of that operation is just to select a range of columns and the row filtering is a syntacitcal detail. We could rewrite this too "The `[]` operation is only used when you want to either filter rows [using a boolean expression] **or** select [a list of columns], but I am not sure if that is too specific.

> also complicated by the fact that you can use [] for str.startswith etc

That was a mistake on my part, you can't do that unless you add an extra step of filtering the column names using the boolean array returned from `str.startswith` so I updated that section to use `loc` instead.

joelostblom:
@trevorcampbell a ping here just because I have started on the first worksheet and tutorial and some of the changes there depend on which decision we take here

trevorcampbell:
OK, I will look at this today at some point!

PR #99, opened 2023-01-05
Ch1 fig cleanup

lheagy:
PR for the code annotation figures in Chapter 1. I also included the ppt where I made the annotations so updates can be made in the future. I also made a couple of minor updates to have the large tables scroll

There is one lingering note for @joelostblom 
image


source/intro.md
```diff
@@ -324,6 +322,7 @@ Syntax for the `read_csv` function
 
 +++
 ```{code-cell} ipython3
+:tags: ["output_scroll"]
```
lheagy, line 325: this is handy for large tables

joelostblom (commented):
I think this looks good overall! What do you think about removing the assignments in the three figures `altair_syntax`, `filter_rows_and_columns`,  and `sort_values`? That will give more real estate for the important part of those images (the explaining of the different parts of the commands)

joelostblom:
I added my section in https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/101

trevorcampbell:
> I think this looks good overall! What do you think about removing the assignments in the three figures altair_syntax, filter_rows_and_columns, and sort_values? That will give more real estate for the important part of those images (the explaining of the different parts of the commands)

I don't think that's worth the edit honestly ;)

This looks good to me aside from the `~$blah.pptx` file that got committed. I'll remove that and merge.

PR #137, opened 2023-01-13
Don't explicitly important altair_saver

joelostblom:
I just noticed this when looking around for a large data section in the viz chapter. We should not need to explicitly important altair saver like that, and we are not actually using that function anyways when doing `chart.save`. Let me know if there was a reason for doing this that I'm missing.


trevorcampbell:
>  Let me know if there was a reason for doing this that I'm missing.

Because I am no expert in altair :) 

I'll review + merge later when I'm going through the final pass of 6/7/8.

trevorcampbell (approved):
LGTM

PR #140, opened 2023-01-18
Another merge...

trevorcampbell:


trevorcampbell:
I'm not sure what's going on here...for some reason github thinks production is behind main by 25 commits, but I just merged `main -> production` so they should be up to date. Locally my git agrees that they're up to date.

joelostblom:
I haven't looked at it closely, but would it help to rebase production on main and force push?

trevorcampbell:
@joelostblom I don't think so, since `git` locally on my machine thinks everything is totally fine as-is.

I say we just leave it be for now -- everything seems fine on the published website. Next time we need to push to production we can handle it.

PR #141, opened 2023-01-23
Chapter 9 production polish

joelostblom:
Not too much in this chapter. Are we changing ggplot plots to altair or leaving as is? We could make 9.10 an interactive figure instead of faceted to make it a bit less busy if we want, but I think it is good as is too.

### DIFF
--- a/source/clustering.md
+++ b/source/clustering.md
@@ -169,8 +169,14 @@ Now we can load and preview the data.
 ```{code-cell} ipython3
 :tags: ["remove-cell"]
 from myst_nb import glue
-data = pd.read_csv('data/toy_penguins.csv')
-data
+import pandas as pd
+
+data = pd.read_csv(
+    'data/toy_penguins.csv'
+).replace(
+    [2, 3],
+    [0, 2]
+)
 ```
 
 ```{code-cell} ipython3
--- a/source/clustering.md
+++ b/source/clustering.md
@@ -15,27 +15,14 @@ kernelspec:
 (clustering)=
 # Clustering
 
-
-```{code-cell} ipython3
-
-import pandas as pd
-import numpy as np
-import random
-from sklearn.preprocessing import StandardScaler
-from sklearn.compose import make_column_transformer
-
-import altair as alt
-
-```
-
 ```{code-cell} ipython3
 :tags: ["remove-cell"]
 import warnings
 warnings.filterwarnings("ignore", category=FutureWarning)
 ```
 
-
 ## Overview
+
 As part of exploratory data analysis, it is often helpful to see if there are
 meaningful subgroups (or *clusters*) in the data.
 This grouping can be used for many purposes,
--- a/source/clustering.md
+++ b/source/clustering.md
@@ -16,7 +16,8 @@ kernelspec:
 # Clustering
 
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
+
 import warnings
 warnings.filterwarnings("ignore", category=FutureWarning)
 ```
@@ -167,7 +168,8 @@ np.random.seed(149)
 Now we can load and preview the data.
 
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
+
 from myst_nb import glue
 import pandas as pd
 
@@ -197,11 +199,11 @@ scatter_plot = alt.Chart(penguin_data).mark_circle().encode(
     x=alt.X("flipper_length_standardized", title="Flipper Length (standardized)"),
     y=alt.Y("bill_length_standardized", title="Bill Length (standardized)")
 )
-
 ```
 
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
+
 glue('scatter_plot', scatter_plot, display=True)
 ```
 
@@ -244,11 +246,8 @@ This procedure will separate the data into groups;
 {numref}`colored_scatter_plot` shows these groups
 denoted by colored scatter points.
 
-
-
-
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
 
 colored_scatter_plot = alt.Chart(data).mark_circle().encode(
     x=alt.X("flipper_length_standardized", title="Flipper Length (standardized)"),
@@ -289,7 +288,8 @@ have.
 ### Measuring cluster quality
 
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
+
 clus = data[data["cluster"] == 2].loc[:,["bill_length_standardized", "flipper_length_standardized"]]
 ```
 
@@ -316,7 +316,7 @@ Then we would compute the coordinates, $\mu_x$ and $\mu_y$, of the cluster cente
 $\mu_x = \frac{1}{4}(x_1+x_2+x_3+x_4) \quad \mu_y = \frac{1}{4}(y_1+y_2+y_3+y_4)$
 
 ```{code-cell} ipython3
-:tags: ["remove-cell"]
+:tags: [remove-cell]
 
 clus_rows = clus.shape[0]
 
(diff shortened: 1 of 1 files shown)


source/clustering.md
```diff
@@ -678,95 +654,60 @@ The data colored by the cluster assignments returned by K-means.
...
+# Create an empty list
+inertias = []
 
-If we take a look at our data frame `penguin_clust_ks` now,
-we see that it has two columns: one with the value for K,
-and the other holding the clustering model objects.
+ks = range(1, 10)
+for k in ks:
+    # Save the computed inertia for each k
+    inertias.append(KMeans(n_clusters=k).fit(standardized_data).inertia_)
```
joelostblom: I think this is more straight-forward than the current approach with apply. Since we already are teaching for loops in chapter 10 I thought it made sense to have one here too. We can of course discuss removing them at both places if we don't think we need them.

source/clustering.md
```diff
@@ -785,61 +726,27 @@ A plot showing the total WSSD versus the number of clusters.
...
+
+> **Note:** It is rare that the KMeans function from scikit-learn
+> gets stuck in a bad solution,
+> because the selection of the centroid starting points
+> is optimized to prevent this from happening.
+> If you still find yourself in a situation where you have a bump in the elbow plot,
+> you can increase the `n_init` parameter above the default value of 10
+> to try more different starting points for the centroids.
+> The larger the value the better from an analysis perspective,
+> but there is a trade-off that doing many clusterings could take a long time.
```
joelostblom: Since the default in sklearn is that everything works as expected, I thought it was better to move this to a note instead of starting with an odd formulation of the params that we wouldn't actually use often.

source/clustering.md
```diff
@@ -290,9 +274,9 @@ where we can easily visualize the clusters on a scatter plot, we can give
 human-made labels to the groups using their positions on
 the plot:
 
-- small flipper length and small bill length (<font color="#D55E00">orange cluster</font>),
-- small flipper length and large bill length (<font color="#0072B2">blue cluster</font>).
-- and large flipper length and large bill  length (<font color="#F0E442">yellow cluster</font>).
+- small flipper length and small bill length (<font color="#4c78a8">blue cluster</font>),
```
trevorcampbell: These labels are not correct I don't think -- randomness changed

source/clustering.md
```diff
@@ -585,9 +569,13 @@ using `fit_transform()`
 
 
 ```{code-cell} ipython3
+from sklearn.preprocessing import StandardScaler
+
 scaler = StandardScaler()
 standardized_data = pd.DataFrame(
-    scaler.fit_transform(not_standardized_data), columns = ['bill_length_mm', 'flipper_length_mm'])
+    scaler.fit_transform(not_standardized_data),
```
trevorcampbell: See the new chapters 5,6,7,8 and try to make the use of StandardScaler / preprocessing follow the same pattern as there

we use make_column_transformer there if i recall

source/clustering.md
```diff
@@ -598,38 +586,31 @@ standardized_data
...
+In machine learning terms,
+we usually say that we "label" each data points
+as belonging to one of these clusters,
+and these labels are contained in the `labels_` attribute of the clustering object.
 
 ```{code-cell} ipython3
-predictions = penguin_clust.predict(standardized_data)
-predictions
-
+labels = penguin_clust.labels_
```
trevorcampbell, line 613: this does not print -- should make it print. otherwise it's not clear what it is

source/clustering.md
```diff
@@ -678,95 +654,60 @@ The data colored by the cluster assignments returned by K-means.
...
-As mentioned above, we also need to select K by finding
-where the "elbow" occurs in the plot of total WSSD versus the number of clusters.
-We can obtain the total WSSD (inertia) from our
-clustering using `.inertia_` function. For example:
+As mentioned above,
+instead of arbitrarily setting K to a number as we did above,
+we can find the best value for K
+by finding where the "elbow" occurs in the plot of total WSSD versus the number of clusters.
+The total WSSD is stored in the `.inertia_` attribute
+of the clustering object ("inertia" is another term for WSSD).
```
trevorcampbell: I have never heard anyone use the name "inertia" -- so I might change this wording to say something like "this is what scikit-learn calls wssd"

source/clustering.md
```diff
@@ -678,95 +654,60 @@ The data colored by the cluster assignments returned by K-means.
...
+To create this dataframe,
+we will use what is called a "for loop"
+where we repeat an operation a number of times.
+Here is an examples of a for loop that simply prints each number from 1-9:
 
 ```{code-cell} ipython3
-import numpy as np
-penguin_clust_ks = pd.DataFrame({"k": np.array(range(1, 10)).transpose()})
+numbers = range(1, 10)
+for number in numbers:
```
trevorcampbell: we don't teach for loops in this class (in fact, we can't! we can talk offline about it if you want)

but to your point in the conversation thread: unfortunately we have to remove for loops from both chapters. 

try to find a reasonable way to accomplish the same thing w/o loops

source/clustering.md
trevorcampbell: see my other comment, but yeah unfortunately we cannot cover for loops

source/clustering.md
```diff
@@ -785,61 +726,27 @@ A plot showing the total WSSD versus the number of clusters.
 ```{index} K-means; init argument
 ```
 
-It looks like 3 clusters is the right choice for this data.
-But why is there a "bump" in the total WSSD plot here?
```
trevorcampbell, line 789: We should keep this example... I'm not sure why it was removed here

You can feel free to seed-hack (in a hidden cell) as much as you like to recreate the example

source/clustering.md
trevorcampbell: > Since the default in sklearn is that everything works as expected

I don't know what that means. Kmeans in sklearn is initialized using Kmeans++, which is not bullet-proof -- you can still end up in local optima. It's just less likely. But it's still important to explain to students (in simpler terms) that Kmeans is a local optimizer, it doesn't find the global optimum and can output different answers each time.

There's no need to replicate the R book exactly (in terms of code), though

Honestly it might be neat to cover Kmeans++ in an optional section (but that's for the future -- no need to do it now. Just open an issue to remind us about the idea later. )

trevorcampbell (changes requested):
Looks pretty good! Some changes requested

trevorcampbell:
> Are we changing ggplot plots to altair or leaving as is? 

I think it's OK to leave as-is for now, but at some point we should re-make them in altair. Open an issue for it please :)

> We could make 9.10 an interactive figure instead of faceted to make it a bit less busy if we want, but I think it is good as is too.

No interactive figs, this is going to go in a hardcopy book at some point most likely so we want to keep things static as much as possible

source/clustering.md
joelostblom: Good catch! Forgot to update them after changing this plot to be consistent with the clustering plot further down.

source/clustering.md
joelostblom: Changed to list comprehension as in ch 10. I think it is more succinct and easier to understand that for loops here even!

source/clustering.md
joelostblom: See my reply to your next comment

source/clustering.md
joelostblom: The point I am trying to make is that with the default sklearn Kmeans param of using Kmeans++ for initialization and 10 runs by default, it is **very** unlikely to get stuck in a local minimum. In fact, I tried seed-hacking with the 10,000 first seed and not a single one resulted in a bump on the chart.

I agree with you that it is important to mention that K-means is a local optimizer, but the example used was contrived and not a realistic scenario of parameter combinations that the students are ever likely to use. That is why I removed the example and still explained the signs of sub-optimal local optimization in the added text.

Of note, scikit-learn is changing the default number of runs in 1.4 to be 1 for kmeans++ (I ran with `n_iter=1` through the 10,000 first seeds here as well and still there was only 3.5% of cases resulting in a bump on the chart, but possibly when we updrade to 1.4 we could seedhack our way to a bump):

image

Code for reference

```py
sum([
    (pd.Series([
        KMeans(n_clusters=k, random_state=seed, n_init=1).fit(standardized_data).inertia_
        for k in range(1, 10)
    ]).diff() > 0).any()
    for seed in range(10_000)
])
```

joelostblom (commented):
@trevorcampbell Thanks for the comments, ready for a second look!

trevorcampbell (approved):
I'm still not 100% happy with removing the bump example -- I don't buy your arguments regarding kmeans++. If it can happen 3.5% of the time when initializing once, that's absolutely frequent enough that I'd want to cover it. Even just keeping this in sync with the R version is enough of a reason for me.

But it's not worth delaying the PR for it. I may go in and edit it later on a second global pass through the book.

PR #142, opened 2023-01-24
Chapter 10 production polish

joelostblom:
Phew, this chapter was definitely more work! The details of each change are in the commit messages. The big ones are changing to use list comprehensions instead of for loops, removing unnecessary resampling to speed things up, being consistent in using sample instead of switching to resample, using value_counts instead of manual calculations, and big simplifications to the visualization syntax. I also simplified quite a bit and removed a fair amount of verbose code.

### DIFF
--- a/source/inference.md
+++ b/source/inference.md
@@ -1187,7 +1187,10 @@ To calculate a 95\% percentile bootstrap confidence interval, we will do the fol
 2. Find the value such that 2.5\% of observations fall below it (the 2.5\% percentile). Use that value as the lower bound of the interval.
 3. Find the value such that 97.5\% of observations fall below it (the 97.5\% percentile). Use that value as the upper bound of the interval.
 
-To do this in Python, we can use the `percentile()` function from the `numpy` package:
+To do this in Python, we can use the `quantile` function of our DataFrame.
+Quantiles are expressed in proportions rather than percentages,
+so the 2.5th and 97.5th percentiles
+would be quantiles 0.025 and 0.975, respectively.
 
 ```{index} numpy; percentile, pandas.DataFrame; df[]
 ```
--- a/source/inference.md
+++ b/source/inference.md
@@ -527,29 +527,28 @@ sample_estimates = (
 sample_estimates
 ```
 
+Now we can calculate the sample mean for each replicate and plot the sampling
+distribution of sample means for samples of size 40.
+
 ```{code-cell} ipython3
 :tags: [remove-output]
 
-sampling_distribution_40 = (
-    alt.Chart(sample_estimates)
-    .mark_bar()
-    .encode(
-        x=alt.X(
-            "sample_mean",
-            bin=alt.Bin(maxbins=30),
-            title="Sample mean price per night (Canadian dollars)",
-        ),
-        y=alt.Y("count()", title="Count"),
-    )
+sampling_distribution = alt.Chart(sample_estimates).mark_bar().encode(
+    x=alt.X(
+        "sample_mean",
+        bin=alt.Bin(maxbins=30),
+        title="Sample mean price per night (Canadian dollars)",
+    ),
+    y=alt.Y("count()", title="Count"),
 )
 
-sampling_distribution_40
+sampling_distribution
 ```
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
-glue("fig:11-example-means4", sampling_distribution_40)
+glue("fig:11-example-means4", sampling_distribution)
 ```
 
 :::{glue:figure} fig:11-example-means4
@@ -561,8 +560,8 @@ Sampling distribution of the sample means for sample size of 40.
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
-glue("quantile_1", round(int(sample_estimates["sample_mean"].quantile(0.25)), -1))
-glue("quantile_3", round(int(sample_estimates["sample_mean"].quantile(0.75)), -1))
+glue("quantile_1", round(int(sample_estimates["sample_mean"].quantile(0.25)), - 1))
+glue("quantile_3", round(int(sample_estimates["sample_mean"].quantile(0.75)), - 1))
 ```
 
 ```{index} sampling distribution; shape
@@ -603,54 +602,28 @@ was \$`r round(mean(airbnb$price),2)`.
 ```{code-cell} ipython3
 :tags: [remove-input]
 
-(
-    (
-        alt.Chart(airbnb, title="Population")
-        .mark_bar(clip=True)
-        .encode(
+glue(
+    "fig:11-example-means5",
+    alt.vconcat(
+        population_distribution.mark_bar(clip=True).encode(
             x=alt.X(
                 "price",
                 bin=alt.Bin(maxbins=30),
                 title="Price per night (Canadian dollars)",
-                axis=alt.Axis(values=list(range(50, 601, 50))),
-                scale=alt.Scale(domain=(min(airbnb["price"]), 600)),
-            ),
-            y=alt.Y("count()", title="Count"),
-        )
-        .properties(width=400, height=150)
-    )
-    & (
-        alt.Chart(one_sample, title="Sample (n = 40)")
-        .mark_bar(clip=True)
-        .encode(
-            x=alt.X(
-                "price",
-                bin=alt.Bin(maxbins=30),
-                title="Price per night (Canadian dollars)",
-                axis=alt.Axis(values=list(range(50, 601, 50))),
-                scale=alt.Scale(domain=(min(airbnb["price"]), 600)),
-            ),
-            y=alt.Y("count()", title="Count"),
-        )
-        .properties(width=400, height=150)
-    )
-    & (
-        alt.Chart(
-            sample_estimates,
-            title="Sampling distribution of the mean for samples of size 40",
-        )
-        .mark_bar(clip=True)
-        .encode(
-            x=alt.X(
-                "sample_mean",
-                bin=True,
-                title="Sample mean price per night (Canadian dollars)",
-                axis=alt.Axis(values=list(range(50, 601, 50))),
-                scale=alt.Scale(domain=(min(airbnb["price"]), 600)),
-            ),
-            y=alt.Y("count()", title="Count"),
-        )
-        .properties(width=400, height=150)
+                scale=alt.Scale(domainMax=700)
+            )
+        ).properties(
+            title='Population', height=150
+        ),
+        sample_distribution.properties(title="Sample (n = 40)").properties(height=150),
+        sampling_distribution.properties(
+            title=alt.TitleParams(
+                "Sampling distribution of the mean",
+                subtitle="For 20,000 samples of size 40"
+            )
+        ).properties(height=150)
+    ).resolve_scale(
+        x='shared'
     )
 )
 ```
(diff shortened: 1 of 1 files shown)


source/inference.md
```diff
@@ -41,6 +39,7 @@ populations and then introduce two common techniques in statistical inference:
 *point estimation* and *interval estimation*.
 
 ## Chapter learning objectives
+
 By the end of the chapter, readers will be able to do the following:
 
 * Describe real-world examples of questions that can be answered with statistical inference.
```
joelostblom, line 45: We don't say anything about sd in the chapter but it is mentioned in the line below

source/inference.md
```diff
@@ -283,45 +255,23 @@ expect our sample proportions from this population to vary for samples of size 4
...
-np.random.seed(1)
-
-samples = []
-for rep in range(20000):
-    sample = airbnb.sample(40)
-    sample = sample.assign(replicate=rep)
-    samples.append(sample)
-samples = pd.concat([samples[i] for i in range(len(samples))])
-
+samples = pd.concat([airbnb.sample(40).assign(replicate=n) for n in range(20_000)])
```
joelostblom, line 268: This is what I suggest we use instead of the loop. I considered making the replicate numbers start at 1 as in the R book, but decided it was unnecessary visual noise just to make it more like R.

source/inference.md
```diff
@@ -333,33 +283,14 @@ Both the first and last few entries of the resulting data frame are printed
...
+    .reset_index(name='sample_proportion')
+    .query('room_type=="Entire home/apt"')
 )
-
-# calculate the proportion
-sample_estimates["sample_proportion"] = sample_estimates["counts"] / 40
-
-# drop the count column
-sample_estimates = sample_estimates.drop(columns=["counts"])
-
```
joelostblom: While I find that this section is cleaner than it used to be, I wish it could be even simpler. Especially the index collision and need to use `name=` is unfortunate since I don't think we mention that in the wrangling chapter, so we might want to add a sentence if we stick with it here.

source/inference.md
```diff
@@ -1468,64 +1170,37 @@ To do this in Python, we can use the `percentile()` function from the `numpy` pa
...
+    ),
+    alt.Chart().mark_rule(color='#f58518', size=3, strokeDash=[5]).encode(x=alt.datum(ci_bounds[0.975])),
+    alt.Chart().mark_text(color='#f58518', size=12, fontWeight='bold').encode(
+        x=alt.datum(ci_bounds[0.975]),
+        y=alt.value(-10),
+        text=alt.datum(f'97.5th percentile ({ci_bounds[0.975].round(1)})')
+    ),
+    width=500
 )
 ```
```
joelostblom, line 1274: The additional resources at the end of the chapter are all about R books..

source/inference.md
trevorcampbell: I like it. I also think it's unnecessary to have "1 - 20,000" like in R. 0 - 19,999 is fine.

source/inference.md
```diff
@@ -333,33 +283,14 @@ Both the first and last few entries of the resulting data frame are printed
...
-    .groupby("replicate")["room_type"]
-    .count()
-    .reset_index()
-    .rename(columns={"room_type": "counts"})
+    samples
+    .groupby('replicate')
+    ['room_type']
+    .value_counts(normalize=True)
+    .reset_index(name='sample_proportion')
+    .query('room_type=="Entire home/apt"')
```
trevorcampbell: I don't think we use `query` much anywhere in the book, and we only really briefly mention it in wrangling. I would prefer one of our more common ways of filtering here if possible

source/inference.md
```diff
@@ -333,33 +283,14 @@ Both the first and last few entries of the resulting data frame are printed
...
-    samples.query("room_type == 'Entire home/apt'")
-    .groupby("replicate")["room_type"]
-    .count()
-    .reset_index()
-    .rename(columns={"room_type": "counts"})
+    samples
+    .groupby('replicate')
+    ['room_type']
+    .value_counts(normalize=True)
+    .reset_index(name='sample_proportion')
```
trevorcampbell, line 332: This use of `reset_index` needs to be explained. It would be nice if it were not used at all, but for a first version I'm OK with it as long as it's described well

source/inference.md
trevorcampbell: yep see my earlier comment

source/inference.md
```diff
@@ -397,19 +328,19 @@ Sampling distribution of the sample proportion for sample size 40.
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
-glue("sample_propotion_center", round(sample_estimates["sample_proportion"].mean(), 1))
-glue("sample_propotion_min", round(sample_estimates["sample_proportion"].min(), 1))
-glue("sample_propotion_max", round(sample_estimates["sample_proportion"].max(), 1))
+glue("sample_proportion_center", round(sample_estimates["sample_proportion"].mean(), 1))
```
trevorcampbell: I think the rounding here is too aggressive -- it says the plot is "centered around 0.7" but to me it looks more like 0.78 or so.

Also I would say not to use the min/max here for the upper/lower bound -- I would say like 5th / 95th or 10th / 90th percentile would be better

source/inference.md
```diff
@@ -1276,7 +1024,7 @@ our point estimate to behave if we took another sample.
 
 ```{code-cell} ipython3
 boot20000_means = boot20000.groupby("replicate")["price"].mean().reset_index().rename(
```
trevorcampbell: probably worth explaining this line more carefully (and formatting it better)

source/inference.md
```diff
@@ -1468,64 +1170,37 @@ To do this in Python, we can use the `percentile()` function from the `numpy` pa
 ```
 
 ```{code-cell} ipython3
-import numpy as np
-bounds = np.percentile(boot20000_means["mean"], [2.5, 97.5])
-
-bounds
+ci_bounds = boot20000_means["sample_mean"].quantile([0.025, 0.975])
```
trevorcampbell, line 1241: The text above this needs to be updated to be about the `quantile` method instead of percentile, and we should probably also point out the naming difference

trevorcampbell (changes requested):
Very nice! minor changes requested

Also, one more thing that I couldn't point out in my review: at the end of the chapter in the Exercises section, make sure to fix the "worksheets repository" link to point to the python stuff. You can get the URL from any of the earlier chapters I edited

source/inference.md
ivanistheone: So the original is:
```Python
samples = []
for rep in range(20000):
    sample = airbnb.sample(40)
    sample = sample.assign(replicate=rep)
    samples.append(sample)
samples = pd.concat([samples[i] for i in range(len(samples))])
```

and the replacement is very good:

```Python
samples = pd.concat([airbnb.sample(40).assign(replicate=n) for n in range(20_000)])
```

but could we maybe split onto two lines as in:

```Python
rows = [airbnb.sample(40).assign(replicate=n) for n in range(20_000)]
samples = pd.concat(rows)
```

While I was thinking about this, I wondered if for loops are not allowed,
then maybe we can simplify the accidental complexity by using a function:

```Python
def sample(data, n):
    return data.sample(n)
samples_list = [sample(data, n).assign(replicate=rep) for rep in range(20000)]
samples = pd.concat(samples_list)
```

and while we're at it, maybe make the function do the whole infer-like chunk-of-dataframe generation:

```Python
def getsample(data, n, rep):
    sample = data.sample(n)
    samplerows = sample.assign(replicate=rep)
    return samplerows
samples = pd.concat(getsample(data, n, rep) for rep in range(20000))
```


To be clear, I think the current re-write is fine, just throwing ideas out there for possible simplifications.

ivanistheone: Oh yeah and Hello, I'm Ivan. I recently learned about he UBC data science courses and I think y'all are doing a great job. I'm particularly interested in the Python parts (bcs I'm only beginner in R). I was following this repo and thought I'd jump in to offer some suggestions. 

I'm dealing with similar issues around using Python to teach stats using the modern curriculum (simulation, permutations tests, and bootstrap estimation), but also not trying to teach all of Python (because students don't identify as "coders" so we don't want to alienate them from learning stats).

I'm starting to think learning Python is inevitable though.... It seems to me that it will take longer to teach learner X statistics directly, than to offer the same learner a [Python tutorial](https://nobsstats.com/tutorials/python_tutorial.html), after which they can learn [modern statistics](https://minireference.com/blog/wp-content/uploads/2021/03/TheIntroductoryStatisticsCourse__Cobb.pdf) based on computational methods based on Python+pandas+scipy.stats+viz(alt,sns,plotnine).

source/inference.md
joelostblom: Agree, I reformatted this and added an explanation in the previous code cell which is where this occurs for the first time.

source/inference.md
joelostblom: Hello @ivanistheone, thank your for taking the time to chime in here. We don't teach functions in this course we can't use those suggestions unfortunately. Regarding breaking up into two steps, I think that is a good idea when we first introduce list comprehensions in the previous chapter, which I am going to review now and keep your suggestion in mind if there is a similar situation occurring there (I also added a sentence to clarify here). I wouldn't personally break it down in two steps on such as big comprehension to avoid storing the data twice, but I think it is good in a toy example for educational purposes.

Thanks again for your comments and kind words about the UBC data science courses!

joelostblom (commented):
@trevorcampbell Thanks for the comments, ready for another look!

source/inference.md
joelostblom: I couldn't find a good way to do this that didn't include something new such as value counts on  a data frame instead of series, or using `as_index=False` with groupby (both also would include an extra rename step). The current behavior is [listed asa  bug with a pending PR in pandas](https://github.com/pandas-dev/pandas/issues/49497), so in 2.0 it will work without the `name=` in the way we have written it now.

source/inference.md
trevorcampbell: +1 -- thanks for your input @ivanistheone !

source/inference.md
```diff
@@ -1257,26 +1052,44 @@ Histograms of first six replicates of bootstrap samples.
...
 
 ```{code-cell} ipython3
-six_bootstrap_samples.groupby("replicate")["price"].mean().reset_index().rename(
-    columns={"price": "mean"}
+(
+    six_bootstrap_samples
+    .groupby("replicate")
+    ["price"]
+    .mean()
+    .rename(columns={"price": "sample_mean"})
```
trevorcampbell: @joelostblom I'm seeing a stack trace here and another nearby (which then causes more downstream errors)

Please fix + verify and let me know!

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[39], line 2
      1 (
----> 2     six_bootstrap_samples
      3     .groupby("replicate")
      4     ["price"]
      5     .mean()
      6     .rename(columns={"price": "sample_mean"})
      7     .reset_index()
      8 )

TypeError: Series.rename() got an unexpected keyword argument 'columns'
```

trevorcampbell: (aside from that the PR looks good!)

trevorcampbell (changes requested):
One remaining bugfix needed

source/inference.md
joelostblom: Woops, I swapped the two last lines and forgot to swap them back, fixed.

trevorcampbell (approved):

PR #157, opened 2023-02-02
Replace "model specification" wording

### LINKED ISSUE #84: Rename "model specification" to "model object"
In Python, there isn't really a "model specification" any more -- we just obtain and manipulate model objects from sklearn.

I will do this in Chapter 5 + 6, but we will need to do it also in at least 7+8 (and maybe beyond).

joelostblom:
close #84


trevorcampbell (approved):
LGTM

PR #158, opened 2023-02-02
Replace "len" with "shape[0]" and don't create empty df

### LINKED ISSUE #123: Number of rows with shape instead of len
from @joelostblom : shape is preferred over length on dataframes since it is already computed and in the spirit of using dataframes methods instead of global python functions in general.

I agree. This is something for a global pass after the initial round of polishing.

### LINKED ISSUE #124: size vs count
E.g. in chapter 5, we currently have a code block like

```
explore_cancer = pd.DataFrame()
explore_cancer['count'] = cancer.groupby('Class')['ID'].count()
explore_cancer['percentage'] = 100 * explore_cancer['count']/len(cancer)
explore_cancer
```

The percentage part could be made a bit cleaner via

```{code-cell} ipython3
100 * cancer.groupby('Class').size() / cancer.shape[0]
```

(note that we replace len with shape also. See #123  )

joelostblom:
close #123 close #124. I left all the occurrences of `len` in cells that students don't see.

### DIFF
--- a/source/classification1.md
+++ b/source/classification1.md
@@ -262,10 +262,7 @@ We have
 tumor observations.
 
 ```{code-cell} ipython3
-explore_cancer = pd.DataFrame()
-explore_cancer['count'] = cancer.groupby('Class')['ID'].count()
-explore_cancer['percentage'] = 100 * explore_cancer['count'] / cancer.shape[0]
-explore_cancer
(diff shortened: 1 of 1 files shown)


trevorcampbell (approved):
All looks good. Surprised I didn't see more uses of `len(`, but searching through I guess they're mostly in hidden cells.

PR #160, opened 2023-02-15
Fixes to visualizations throughout the book

### LINKED ISSUE #57: Better explanation of alt.X and alt.Y
Take from 531 lecture notes.

### LINKED ISSUE #128: Simplify plots
For after our initial polish round: the MDS students tried to replicate the plots of the R version (down to fairly detailed stylistic stuff), but we are probably better off just sticking with Altair defaults for the most part as long as they're reasonable. This will keep the plotting code simple and unobtrusive.

joelostblom:
@lheagy @trevorcampbell I had some free time and started fixing some of the open viz issues we have and noticed a few things along the way, so I did a full review on the viz sections of the entire book. Although this PR includes quite a lot of changes, most are clarifying in nature, either making the explanations more intuitive or slightly changing how we create a viz. I also removed any leftovers from the ggplot translation to make it easier for students when they see altair code online such as using the documentation. Lastly I standardized the syntax in all chapters to what was used in the viz chapter so that we don't mix different types of syntax throughout the book. The commit messages include more details about the changes.

I had a quick look at the worksheets and I think the only thing that would need updating there is the scaffold code that students use to match with the syntax in the viz chapter. I can have a more throughout review of the worksheets after you have reviewed this PR, but no hurry.

close #57 close #128


trevorcampbell:
Great stuff @joelostblom -- I imagine this is one for the end of semester (since the course is in full swing at this point)

joelostblom:
Yeah probably, at least not merging it to production before the semester is over for sure.

source/classification2.md
```diff
@@ -330,16 +330,11 @@ cancer['Class'] = cancer['Class'].replace({
 # create scatter plot of tumor cell concavity versus smoothness,
```
trevorcampbell, line 330: The axes on Figure 6.3 look off to me -- it's very squished horizontally. Seems like perhaps the plot is trying to force include the origin for some reason?

trevorcampbell: (I dont think this is related to a change you made though, but something we should fix in this PR)

source/regression2.md
```diff
@@ -274,7 +274,7 @@ for i in range(len(slope_l)):
             )
         )
         .mark_line(color=line_color_l[i])
-        .encode(x=alt.X("x"), y=alt.Y("y"))
+        .encode(x="x", y="y")
```
trevorcampbell, line 277: We should be as consistent as possible everywhere -- do we use `(x="x", y="y")` everywhere, or `alt.X("x")` etc everywhere? should pick one or the other and stick with it, ideally, assuming that's possible

source/viz.md
```diff
@@ -85,7 +85,7 @@ As with most coding tasks, it is totally fine (and quite common) to make
 mistakes and iterate a few times before you find the right visualization for
 your data and question. There are many different kinds of plotting
 graphics available to use (see Chapter 5 of *Fundamentals of Data Visualization* {cite:p}`wilkeviz` for a directory).
-The types of plot that we introduce in this book are shown in {numref}`plot_sketches`;
+The types of plots that we introduce in this book are shown in {numref}`plot_sketches`;
```
trevorcampbell, line 88: the original text here is grammatically correct -- should be "types of plot" not "types of plots"

source/viz.md
```diff
@@ -255,32 +256,38 @@ Scatter plots show the data as individual points with `x` (horizontal axis)
...
-	  choose a reasonable lower bound based on our data, and that lower bound
-	  doesn't have to be zero.
+    - Here, we set the `x` axis of the plot to the `date_measured` variable,
+      and on the `y` axis, we plot the `ppm` variable.
+    - For the y-axis, we also provided the argument
+      `scale=alt.Scale(zero=False)`. By default, `altair` chooses the y-limits
+      based on the data and will keep `y=0` in view.
+      This is often a helpful default, but here it makes it
+      difficult to see any trends in our data since the smallest value is >300
+      ppm. So by providing `scale=alt.Scale(zero=False)`, we tell altair to
```
trevorcampbell, line 282: is "altair" typeset using code or text? `altair` vs altair. We should be consistent throughout. I believe in the R book, packages are typeset using code syntax, so here we would prefer `altair`, if I recall

trevorcampbell: (this will inevitably be a formatting pass that we have to do later on though)

source/viz.md
```diff
@@ -405,26 +412,21 @@ visualization slightly. Note that it is totally fine to use a small number of
 visualizations to answer different aspects of the question you are trying to
 answer. We will accomplish this by using *scale*,
 another important feature of `altair` that easily transforms the different
-variables and set limits.  We scale the horizontal axis using the `alt.Scale(domain=['1990', '1993'])` by restricting the x-axis values between 1990 and 1994,
-and the vertical axis with the `alt.Scale(zero=False)` function, to not start the y-axis with zero.
+variables and set limits.
 In particular, here, we will use the `alt.Scale` function to zoom in
-on just five years of data (say, 1990-1994). The
+on just a few years of data (say, 1990-1995). The
```
trevorcampbell, line 417: is this an error we made in the R book too?

source/viz.md
```diff
@@ -504,14 +504,11 @@ neither of the variables here have a natural order to them.
 So a scatter plot is likely to be the most appropriate
 visualization. Let's create a scatter plot using the `altair`
 package with the `waiting` variable on the horizontal axis, the `eruptions`
-variable on the vertical axis, and the `mark_point` geometric object.
-By default, `altair` draws only the outline of each point. If we would
-like to fill them in, we pass the argument `filled=True` to `mark_point`. In
-place of `mark_point(filled=True)`, we can also use `mark_circle`.
+variable on the vertical axis, and `mark_point` as the graphical mark.
```
trevorcampbell, line 507: I gotta say: it's totally insane to me that altair treats "point" as "not-filled-in circle", and "circle" as "filled disk". I think the vast majority of people would consider a "point" to be a "dot" or "small filled circle", and "circle" to mean "not-filled-in circle". Blech

source/viz.md
```diff
@@ -609,14 +606,22 @@ can_lang
...
+can_lang = can_lang[(can_lang['most_at_home'] > 0) & (can_lang['mother_tongue'] > 0)]
 ```
 
 ```{index} altair; mark_circle
 ```
 
 We will begin with a scatter plot of the `mother_tongue` and `most_at_home` columns from our data frame.
-The resulting plot is shown in {numref}`can_lang_plot`
+As you have seen in the scatter plots in the previous section
+the default behavior of `mark_point` is to draw the outline of each point.
```
trevorcampbell, line 617: "in the previous section," <-- add comma

source/viz.md
```diff
@@ -609,14 +606,22 @@ can_lang
...
+If we would like to fill them in,
```
trevorcampbell, line 618: awkward switch from "you" to "we"

source/viz.md
```diff
@@ -770,6 +774,45 @@ glue('can_lang_plot_log', can_lang_plot_log, display=False)
 Scatter plot of number of Canadians reporting a language as their mother tongue vs the primary language at home with log adjusted x and y axes.
 :::
 
+You will notice two things in the chart above,
+changing the axis to log creates many axis ticks and gridlines,
+which makes the appearance of the chart rather noisy
+and it is hard to focus on the data.
+You can also see that the second last tick label is missing on the x-axis;
+Altair dropped it because there wasn't space to fit in all the large numbers next to each other.
```
trevorcampbell, line 782: Altair vs `altair` vs altair like above

source/viz.md
```diff
@@ -770,6 +774,45 @@ glue('can_lang_plot_log', can_lang_plot_log, display=False)
...
+:tags: ["remove-cell"]
+glue('can_lang_plot_log_revised', can_lang_plot_log_revised, display=False)
+```
+
+:::{glue:figure} can_lang_plot_log_revised
+:figwidth: 700px
+:name: can_lang_plot_log_revised
+
+Scatter plot of number of Canadians reporting a language as their mother tongue vs the primary language at home with log adjusted x and y axes. Only the major gridlines are shown and the suffix "k" indicate 1,000 ("kilo"), while the suffix "M" indicates 1,000,000 ("million").
+:::
```
trevorcampbell, line 815: "log adjusted" -> "log-adjusted"

"are shown and the" -> "are shown. The"  <--- separate into two sentences

"indicate" -> "indicates"  typo

source/viz.md
```diff
@@ -805,23 +848,23 @@ you can clearly see the mutated output from the table.
 ```
 
 ```{code-cell} ipython3
+canadian_population = 35_151_728
```
trevorcampbell, line 855: this is a nice way to format a number, but have we explained to students that we can do this? This should probably appear in an earlier chapter if not...or maybe here if it's the first time it's relevant / used

source/viz.md
```diff
@@ -805,23 +848,23 @@ you can clearly see the mutated output from the table.
...
 
-Finally, we will edit the visualization to use the percentages we just computed
+Next, we will edit the visualization to use the percentages we just computed
 (and change our axis labels to reflect this change in
 units). {numref}`can_lang_plot_percent` displays
 the final result.
-
-
+Here all the tick labels fit by default so we are not changing the labels to include suffixes
+(note that suffixes can also be harder to understand for small quantities,
```
trevorcampbell: separate long parenthetical out into its own sentence, also edit the new sentence to shorten it a bit for clarity

source/viz.md
```diff
@@ -995,25 +1030,29 @@ Scatter plot of percentage of Canadians reporting a language as their mother ton
...
+the default `altair` color scheme, which is called `'tableau10'`. This is an appropriate choice for most situations and is also easy to read for people with reduced color vision.
+In general, the color schemes that are used by default in Altair are adapted to the type of data that is displayed and selected to be easy to interpret both for people with good and reduced color vision.
+If you are unsure about a certain color combination, you can use
+this [color blindness simulator](https://www.color-blindness.com/coblis-color-blindness-simulator/) to check
+if your visualizations are color-blind friendly.
 
 ```{index} color palette; color blindness simulator
 ```
 
+All the available color schemes and information on how to create your own, can be viewed [in the documentation](https://altair-viz.github.io/user_guide/customization.html#customizing-colors).
```
trevorcampbell: remove comma

trevorcampbell: also "in the altair documentation" <-- specify altair

source/viz.md
```diff
@@ -995,25 +1030,29 @@ Scatter plot of percentage of Canadians reporting a language as their mother ton
...
+To change the color scheme of our chart,
+we can add the `scheme` argument in the `scale` of the `color` encoding.
 Below we pick the `"dark2"` theme, with the result shown
 in {numref}`can_lang_plot_theme`
```
trevorcampbell: missing period after figure reference

source/viz.md
```diff
@@ -1055,14 +1091,63 @@ glue('can_lang_plot_theme', can_lang_plot_theme, display=False)
...
 
-From the visualization in {numref}`can_lang_plot_theme`,
+The chart above gives a good indication of how the different language categories differ,
+and this information is sufficient to answer our research question.
+But what if we want to know exactly which language correspond to which point in the chart?
+With a regular visualization library this would not be possible,
+as adding text labels for each individual language
+would add a lot of visual noise and make the chart difficult to interpret.
+However, since Altair is an interactive visualization library we can add information on demand
+via the `Tooltip` encoding channel,
```
trevorcampbell, line 1106: this is cool, but we need to be very careful with how we include this due to the print version of the book obviously not being able to show interactive figures. 

Is it possible to render a different image in the PDF version of the book that is just a static screenshot of what the tooltip would look like? If not immediately, please open an issue to sort that out (and anywhere else we add interactive figures where it's important that the interactivity is displayed -- though I think it will only matter here in Ch4)

source/viz.md
```diff
@@ -1055,14 +1091,63 @@ glue('can_lang_plot_theme', can_lang_plot_theme, display=False)
...
+:tags: ["remove-cell"]
+# Increasing the dimensions makes all the ticks fit in jupyter book (the fit with the default dimensions in jupyterlab)
+glue('can_lang_plot_tooltip', can_lang_plot_tooltip.properties(height=320, width=420), display=False)
+```
+
+:::{glue:figure} can_lang_plot_tooltip
+:figwidth: 700px
+:name: can_lang_plot_tooltip
+
+Scatter plot of percentage of Canadians reporting a language as their mother tongue vs the primary language at home colored by language category with color-blind friendly colors. Hover over the data points with the mouse pointer to see additional information.
```
trevorcampbell, line 1145: stuff like "hover over things with the mouse" do not look great in a printed / PDF book

trevorcampbell: (but it's good for the web version, hence trying to render two separate copies)

source/viz.md
```diff
@@ -1104,10 +1189,16 @@ Here, we have a data frame of Earth's landmasses,
 and are trying to compare their sizes.
 The right type of visualization to answer this question is a bar plot.
 In a bar plot, the height of the bar represents the value of a summary statistic
-(usually a size, count, proportion or percentage).
+(usually a size, count, sum, proportion or percentage).
```
trevorcampbell: oxford comma pls

source/viz.md
```diff
@@ -1104,10 +1189,16 @@ Here, we have a data frame of Earth's landmasses,
...
 They are particularly useful for comparing summary statistics between different
 groups of a categorical variable.
 
+> **Note:** Although bar charts are often used to display mean and median values,
+> they are generally a poor choice for these types of visualizations and should be avoided.
```
trevorcampbell: I think this nuance is too much for this relatively early chapter and will confuse a lot of our students. I suggest removing it. 

It's a good point though. If we expanded our chapter with boxplots, I think that's where you'd make this argument w/o losing people. perhaps open an issue to add a section on boxplots, copy/paste this text into that issue so we can save it for later, and delete the note box here.

source/viz.md
```diff
@@ -1141,9 +1233,9 @@ question we asked was only about the largest landmasses; let's make the plot a
 little bit clearer by keeping only the largest 12 landmasses. We do this using
 the `nlargest` function; the first argument is the number of rows we want and
 the second is the name of the column we want to use for comparing who is
-largest.  Then to help us make sure the labels have enough
-space, we'll use horizontal bars instead of vertical ones. We do this by
-swapping the `x` and `y` variables.
+largest. Then to help make the landmass labels easier to read
+we'll swap the `x` and `y` variables,
+so that the landmass labels are on the y-axis and we don't have to tilt our head to read them.
```
trevorcampbell: "landmass" isn't a word, it's a variable name -- I would just say "labels" rather than "landmass labels" anyway

trevorcampbell: apparently it is a "newer spelling" and "land mass" is the more traditional. i guess it doesn't matter much, maybe we can just keep landmass

source/viz.md
```diff
@@ -1169,41 +1262,36 @@ Bar plot of size for Earth's largest 12 landmasses.
...
-("Which are the top 7 largest landmasses continents?") in the affirmative.
-But the question could be made clearer from the plot
-by organizing the bars not by alphabetical order
-but by size, and to color them based on whether they are a continent.
-The data for this is stored in the `landmass_type` column.
-To use this to color the bars,
-we use the `color` argument to color the bars according to the `landmass_type`
+The plot in {numref}`islands_bar_top` is clearer now,
+and allows us to answer our question:
+"Which are the top 7 largest landmasses continents?".
```
trevorcampbell: this isn't the question -- the question should be 

"Are the top 7 largest landmasses continents?" 

which is why you need "in the affirmative" after.

source/viz.md
```diff
@@ -1169,41 +1262,36 @@ Bar plot of size for Earth's largest 12 landmasses.
...
+However, we could still improve this visualization
+by organizing the bars by landmass size rather than by alphabetical order.
+We could also color the bars based on whether they are a continent or not
```
trevorcampbell: remove "or not"

trevorcampbell: "they are" -> "they correspond to"

trevorcampbell: here is what i'm changing the R version to read:

image

source/viz.md
```diff
@@ -1169,41 +1262,36 @@ Bar plot of size for Earth's largest 12 landmasses.
...
+in the y-encoding of the chart.
+Since the `size` variable is encoded in the x channel of the chart,
+we specify `sort='x'` inside `alt.Y`.
+This plots the values on `y` axis
 in the ascending order of `x` axis values.
-We do this here so that the largest bar will be closest to the axis line,
-which is more visually appealing. If instead, we want to sort the values on `y-axis` in descending order of `x-axis`, we need to specify `sort='-x'`.
+This creates a chart where the largest bar is the closest to the axis line,
+which is generally the most visually appealing when sorting bars.
+If instead,
```
trevorcampbell: remove comma "If instead"

source/viz.md
```diff
@@ -1219,22 +1307,36 @@ Bar plot of size for Earth's largest 12 landmasses colored by whether its a cont
...
-We can make one more finishing touch in {numref}`islands_plot_titled`: we will
-add a title to the chart by specifying `title` argument in the `alt.Chart` function.
-Note that plot titles are not always required; usually plots appear as part
+making it quite clear that all the seven largest landmasses are continents.
+
+To add some finishing touches to the chart
+we will add a title to the chart by specifying `title` argument inside `alt.Chart`.
+A good plot title should usually contain the take home message that you want readers of the chart to focus on,
+e.g. "The Earth's seven largest landmasses are all continents",
+but it could sometimes be more general, e.g. "The twelve largest landmasses on Earth".
```
trevorcampbell: NB: this stuff about titles seems to have been removed from the R version. I put a pretty high value on syncing the two books as much as possible to avoid divergent material, so we should decide whether this is included or removed or de-synced from R

trevorcampbell: (though obviously there will be differences due to differing syntax / code patterns etc, but the set of "ideas/topics we teach" should flow the same)

trevorcampbell (changes requested):
Amazing improvement @joelostblom ! I have lots of fiddly comments but overall it's a great update.

trevorcampbell:
> I can have a more throughout review of the worksheets after you have reviewed this PR, but no hurry.

Oh also yeah it would be awesome if you could open a PR that syncs the worksheets/tutorials with your new material too :)

(or at least open an issue that specifies what would need to change)

source/viz.md
joelostblom: 100% agree, I'm used to it now, but it was definitely one of the things I reacted to as unintuitive when starting to use the library.

source/classification2.md
joelostblom: Changed from

image


to

image

source/regression2.md
joelostblom: I believe we only use the `alt.X/Y` syntax when setting an option. Although technically we could always use it, even when not setting options, it is often convenient to write just `x='x'` instead of `alt.X('x')`, but I'm not strongly against changing to always use `alt.X/Y` since that would allow us to not specify the named parameter, however I don't think it is an important change and code in the wild + the docs  often uses both syntax so it is good to know of them.

source/viz.md
joelostblom: Hmm, wouldn't this be similar to saying something like "The types of friends I’ve made in this college are great." where both are plural? Or the "The types of apples that I like are red." Both from this article https://grammarhow.com/type-of-types-of-followed-by-singular-or-plural/. This other article suggest that it might be a difference between American and British English https://english.stackexchange.com/a/5546, with the "type of plurals" being AmE which is what we use in the rest of the book. (I'm realizing I'm ESL and might be missing something but got interested and had to look it up =p )

source/viz.md
joelostblom: In the docs, we just call it "Altair" without typesetting as code and with a leading capital. I agree we should change it in a later formatting pass if at all.

source/viz.md
joelostblom: No, it is fine in the R book, but Altair will switch to include the name of the months if we only do 90-94 which is why I extended it with one year

image

versus

image

source/viz.md
joelostblom: Added a one sentence explanation

source/viz.md
joelostblom: Good point, tracking in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/175

source/viz.md
joelostblom: Also in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/175

source/viz.md
joelostblom: Hmm this is what we have in the question further up:

image

Which to me is more similar to "Which of the top 7 largest landmasses are continents?" (fixing a couple of typos)

source/viz.md
joelostblom: Updated based on your screenshot with a slight tweak in the wording to match altair's syntax

source/viz.md
joelostblom: I agree that it would be valuable to keep them in sync and would prefer to have this section added to the R book. I think it is important to mention titles and show examples of what an appropriate title looks like.

joelostblom (commented):
Thanks for the thorough review @trevorcampbell ! I think I have addressed all your comments or opened issues to follow up so please have a look.

source/viz.md
trevorcampbell: honestly it's so minor that I don't mind either way. 

I don't know the actual rule, and I was just parsing it logically: "types of plot" means multiple kinds of plot. "types of plots" means multiple kinds of...multiple plots?

source/viz.md
trevorcampbell: good catch!

source/viz.md
trevorcampbell: We should keep the wording the same throughout if possible -- so "are the continents the largest 7" or the like

(also if the original Q were just "which of the top 7 are continents" I'd just plot 7 instead of whatever we're doing now, 12 or something like that)

trevorcampbell:
@joelostblom thanks for the fixes. I just have that one remaining comment about the wording of the Q. I think we should keep it as consistent as possible throughout the chapter. Don't see a need to change it from what's in the R version and in this one earlier in the chapter

Aside from that, looks good to me!

source/viz.md
joelostblom: I guess it makes it more complicated that our original question is two-fold: both "Are the continents (North / South America, Africa, Europe, Asia, Australia, Antarctica) Earth’s seven largest landmasses?" and then "If so, what are the next few largest landmasses after those?". I updated to include both of those with the same phrasing as in the original question to keep it consistent, let me know what you think.

I believe the reason we have 12 instead of 7 is to show of the color encoding which would just contain one color if we only included the top 7 since those are all continents:

image

source/viz.md
joelostblom: Sorry I missed this comment initially. I think this is more than a nuance and a common mistake I see students making, so I would be in favor of adding it to the R and Python books. I think it could fit nicely as a motivation for the histogram section, which could begin by comparing the average and then showing the actual histograms of all observations instead. What do you think of that?

source/viz.md
trevorcampbell: We have a style guide for the R textbook here: https://github.com/ubc-dsci/introduction-to-datascience#style-guide

we can just follow this (and update it as necessary) once we are at the level of "final polish" for the official publication

source/viz.md
trevorcampbell: I'm still not entirely convinced...

Idea: can we open a separate PR (or an issue) to discuss this change? I don't want to block this PR on that

source/viz.md
```diff
@@ -1170,40 +1268,38 @@ Bar plot of size for Earth's largest 12 landmasses.
...
+and allows us to answer our initial questions:
+"Are the seven continents Earth's largest landmasses?"
+and "Which are the next few largest landmasses?".
+However, we could still improve this visualization
+by organizing the bars by landmass size rather than by alphabetical order
+and by coloring the bars based on whether they correspond to a continent.
 The data for this is stored in the `landmass_type` column.
-To use this to color the bars,
-we use the `color` argument to color the bars according to the `landmass_type`
+TO use this to color the bars,
```
trevorcampbell: @joelostblom typo added here "TO" should be "To"

source/viz.md
```diff
@@ -1170,40 +1268,38 @@ Bar plot of size for Earth's largest 12 landmasses.
...
+we set `color` encoding to `landmass_type`.
```
trevorcampbell: @joelostblom typo "set the"

source/viz.md
trevorcampbell: @joelostblom it took me a bit to dig around in my memory for why we decided against this in R, but I finally remembered: we don't use titles anywhere else in the book. If we're suggesting Ch4 is the guide to making good viz, we should follow our own rules elsewhere (and we do, generally). But we felt strongly that we shouldn't have titles everywhere else in the book -- at a minimum, the publisher wouldn't want them in the print book format. So since we weren't including titles everywhere, we didn't feel it would be good to suggest it in Ch4.

I was originally more on the fence about this, but now I'm a solid "no titles" to avoid causing a huge transitive-closure headache elsewhere

source/viz.md
joelostblom: Opened https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/178

source/viz.md
joelostblom: In the book we do have captions everywhere, which fill a similar need as the titles we are explaining to students and help clarify what is figure contains. I think this is important and we don't want student to produce charts without any explanation (whether it is a title and caption), but removing for now so that we can move forward. Follow up in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/179

joelostblom:
@trevorcampbell Thanks for the comments, I think I have addressed everything now!

trevorcampbell (approved):
It's excellent! Thanks so much for the effort @joelostblom -- a nice improvement here.

Issue #82, opened 2023-01-01
Images being written to `source/`


trevorcampbell:
In Chapter 5, I saw an image being written to the source folder

```
plot(fig, filename="img/classification1/fig05-more.html", auto_open=False)
```

We should go through the whole book and ensure that all output is only being written to `_build/`

trevorcampbell:
I see now why things were set up this way...saving to `_build` does not seem to work.

It's not great though. We should revisit later. This is not super high priority for now.

trevorcampbell:
After #240 , the 3d viz are no longer saved to disk. So this might be stale now. Needs investigation

trevorcampbell:
Confirmed that this is no longer happening. Closing

PR #161, opened 2023-02-15
Change source permissions back to 755 after running the docer container

### LINKED ISSUE #89: Fix permissions on source
For some reason source is set to be writable by other users than the owner and group. I don't think there is any good reason for this so we should set it back to permissions 775 at some point.

image

joelostblom:
Once we figure out a solution to #82, we can avoid writing to source entirely. This is a temporary solution which at least avoids that the committed files on github have the more permissive permissions (we would need to commit these changes to the files at some points regardless of solution). close #89


build_html.sh
```diff
@@ -1,2 +1,3 @@
 chmod -R o+w source/
 docker run --rm -v $(pwd):/home/jovyan ubcdsci/py-intro-to-ds:20230104230634037f38 /bin/bash -c "jupyter-book build source"
+chmod -R 755 source/
```
trevorcampbell: I generally go with the more semantic `o-w` to remove write perms. I'll fix that and merge

PR #169, opened 2023-07-06
Minor edits summer 2023

trevorcampbell:
This PR syncs a bunch of minor changes that were made to the R version in https://github.com/orgs/UBC-DSCI/projects/5/views/1 and https://github.com/UBC-DSCI/introduction-to-datascience/pull/486

The one thing worth mentioning is that I am getting rid of the "getting your files off of JHub" appendix from the book. I don't think the book is the right place to have this material that's super specific to our course. That's something we should post on Canvas or in our slides.


chendaniely (approved):
Did a quick look through all the changed files. Nothing looks out of place

PR #186, opened 2023-07-27
Polishing flow on the Ch3 PR

trevorcampbell:
This is an edit on #97 -- since I made quite a few changes, I decided to open a separate PR that will be merged into that PR itself prior to going into `main`.

### DIFF
--- a/source/wrangling.md
+++ b/source/wrangling.md
@@ -841,16 +841,27 @@ indicating they are integer data types (i.e., numbers)!
 
 Now that the `tidy_lang` data is indeed *tidy*, we can start manipulating it
 using the powerful suite of functions from the `pandas`.
-We revisit the `[]` from the chapter on {ref}`intro`,
-which lets us create a subset of rows from a data frame.
-Recall the argument to `[]`:
-a list of column names, or a logical statement that evaluates to either `True` or `False`,
-where `[]` returns the rows where the logical statement evaluates to `True`.
-This section will highlight more advanced usage of the `[]` function.
-In particular, this section provides an in-depth treatment of the variety of logical statements
+We will first revisit the `[]` from the chapter on {ref}`intro`,
+which lets us obtain a subset of either the rows **or** the columns of a data frame.
+This section will highlight more advanced usage of the `[]` function,
+including an in-depth treatment of the variety of logical statements
 one can use in the `[]` to select subsets of rows.
 
 +++
+
+### Extracting columns by name
+
+Recall that if we provide a list of column names, `[]` returns the subset of columns with those names.
+Suppose we wanted to select the columns `language`, `region`,
+`most_at_home` and `most_at_work` from the `tidy_lang` data set. Using what we
+learned in the chapter on {ref}`intro`, we can pass all of these column 
+names into the square brackets.
+
+```{code-cell} ipython3
+:tags: ["output_scroll"]
+tidy_lang[["language", "region", "most_at_home", "most_at_work"]]
+```
+
 
 ### Extracting rows that have a certain value with `==`
 Suppose we are only interested in the subset of rows in `tidy_lang` corresponding to the
@@ -1022,7 +1033,10 @@ to make long chains of filtering operations a bit easier to read.
 The `[]` operation is only used when you want to either filter rows **or** select columns;
 it cannot be used to do both operations at the same time. This is where `loc[]`
 comes in. For the first example, recall `loc[]` from Chapter {ref}`intro`,
-which lets us create a subset of columns from a data frame.
+which lets us create a subset of the rows and columns in the `tidy_lang` data frame.
+In the first argument to `loc[]`, we specify a logical statement that 
+filters the rows to only those pertaining to the Toronto region, 
+and the second argument specifies a list of columns to keep by name.
 
 ```{code-cell} ipython3
 :tags: ["output_scroll"]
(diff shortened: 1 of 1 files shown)


source/wrangling.md
```diff
@@ -841,17 +841,28 @@ indicating they are integer data types (i.e., numbers)!
...
-In particular, this section provides an in-depth treatment of the variety of logical statements
+We will first revisit the `[]` from the chapter on {ref}`intro`,
+which lets us obtain a subset of either the rows **or** the columns of a data frame.
+This section will highlight more advanced usage of the `[]` function,
+including an in-depth treatment of the variety of logical statements
 one can use in the `[]` to select subsets of rows.
 
 +++
 
+### Extracting columns by name
```
trevorcampbell, line 852: Reasoning for this edit: This is essentially a repeat of material from Ch1. But we do the same in the R version, and it doesn't hurt to have it here, especially if someone is going to search for how to subset rows/cols and end up in Ch 3 (maybe missing Ch 1 entirely). The section is named "extract rows or columns", so a bit odd not to discuss columns at all.

source/wrangling.md
```diff
@@ -1032,53 +1046,61 @@ tidy_lang.loc[
 ]
 ```
 
-### Using `loc[]` to select ranges of columns
-
-Suppose we wanted to select only the columns `language`, `region`,
-`most_at_home` and `most_at_work` from the `tidy_lang` data set. Using what we
-learned in the chapter on {ref}`intro`, we would pass all of these column names into the square brackets.
+In addition to simultaneous subsetting of rows and columns, `loc[]` has two
```
trevorcampbell, line 1049: @joelostblom I think what was missing before was an emphasis on why we would care about loc[]. I've written it here to reference two special abilities of loc beyond [] -- ranges and logical statements to select columns.

source/wrangling.md
```diff
@@ -1032,53 +1046,61 @@ tidy_lang.loc[
...
+more special capabilities beyond those of `[]`. First, `loc[]` has the ability to specify *ranges* of rows and columns. 
+For example, note that the list of columns `language`, `region`, `most_at_home`, `most_at_work` 
+corresponds to the *range* of columns from `language` to `most_at_work`.
+Rather than explicitly listing all of the column names as we did above,
+we can ask for the range of columns `"language":"most_at_work"`; the `:`-syntax 
+denotes a range, and is supported by the `loc[]` function, but not by `[]`.
 
 ```{code-cell} ipython3
 :tags: ["output_scroll"]
-tidy_lang[["language", "region", "most_at_home", "most_at_work"]]
```
trevorcampbell, line 1043: It was weird to have the first example in the "using loc" section be `[]`. Now the first example is a repeat of what we did in Ch 1 (basic usage), followed by "here's what else you can do with loc"

source/wrangling.md
```diff
@@ -1344,15 +1360,17 @@ region_lang.groupby("region")
...
 
 ```{code-cell} ipython3
 region_lang.groupby("region")["most_at_home"].agg(["min", "max"])
 ```
 
 The resulting dataframe has `region` as an index name.
-This is similar to what happened when we reshaped data frames in the previous chapter,
+This is similar to what happened when we used the `pivot` function
+in the section on {ref}`pivot-wider`;
 and just as we did then,
```
trevorcampbell, line 1374: it's not in the previous chapter -- it's in the pivot section

source/wrangling.md
```diff
@@ -1292,12 +1308,12 @@ region_lang.mean(numeric_only=True)
...
 
 If there are only some columns for which you would like to get summary statistics,
-you can first use `[]` to select those columns
-and then ask for the summary statistic,
-as we did for a single column previously:
-Lets say that we want to know
-the mean and standard deviation of all of the columns between `"mother_tongue"` and `"lang_known"`.
-We use `[]` to specify the columns and then `agg` to ask for both the `mean` and `std`.
+you can first use `[]` or `.loc[]` to select those columns,
+and then ask for the summary statistic
```
trevorcampbell, line 1312: your example right after used `.loc[]` so I added that to the text

source/wrangling.md
```diff
@@ -1110,32 +1132,26 @@ has index `1`!).
...
-```{code-cell} ipython3
-tidy_lang.iloc[:10, :]
-```
-
-`pandas` also provides a shorthand for selecting ranges of rows by using `[]`:
+We can also use `iloc[]` to select ranges of rows, or simultaneously select ranges of rows and columns, using a similar syntax.
+For example, to select the first five rows and columns after and including index 1, we could use the following:
 
 ```{code-cell} ipython3
-tidy_lang[:10]
```
trevorcampbell, line 1131: I got rid of this shorthand example entirely. I think it will confuse students with the `[]` operator earlier. They certainly might (will eventually) see it in the wild, but I think the priority here is to make sure everyone is understanding what we teach and to keep things contained, and not necessarily to cover every possible thing they'll see beyond the class

source/wrangling.md
```diff
@@ -1032,53 +1046,61 @@ tidy_lang.loc[
...
 tidy_lang.loc[:, "language":]
 ```
-
 By not putting anything after the `:`, Python reads this as "from `language` until the last column".
+Similarly, we can specify that we want everything up to and including `language` by writing
+the expression:
+
+```{code-cell} ipython3
+:tags: ["output_scroll"]
+tidy_lang.loc[:, :"language"]
```
trevorcampbell, line 1088: I figured including a "before" example would be helpful here too -- this is a bit weird syntax, so being a bit more explicit is useful

trevorcampbell:
@joelostblom this PR edits your other PR -- if you are happy with these edits, we can merge this one and then merge your PR into `main`

source/wrangling.md
joelostblom: I like this additional motivation!

joelostblom (approved):
This looks good to me, thanks for making these changes!

PR #187, opened 2023-08-01
Replace `help` with `?`

joelostblom:
Nicer formatted documentation as per https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/pull/794/files#r2533

### DIFF
--- a/source/intro.md
+++ b/source/intro.md
@@ -1057,18 +1057,17 @@ Bar plot of the ten Aboriginal languages most often reported by Canadian residen
 
 There are many Python functions in the `pandas` package (and beyond!), and
 nobody can be expected to remember what every one of them does
-or all of the arguments we have to give them. Fortunately, Python provides
-the `help` function, which
+or all of the arguments we have to give them. Fortunately, Jupyter provides
 provides an easy way to pull up the documentation for
-most functions quickly. To use the `help` function to access the documentation, you
-just put the name of the function you are curious about as an argument inside the `help` function.
+most functions quickly by using the `?` character. To access the documentation this way, you
+prefix `?` to the name of the function you are curious about.
 For example, if you had forgotten what the `pd.read_csv` function
 did or exactly what arguments to pass in, you could run the following
 code:
 
 ```{code-cell} ipython3
 :tags: ["remove-output"]
-help(pd.read_csv)
+?pd.read_csv
 ```
 
 {numref}`help_read_csv` shows the documentation that will pop up,
@@ -1088,7 +1087,7 @@ you about the different arguments and usage of functions that you have already l
 height: 700px
 name: help_read_csv
 ---
-The documentation for the read_csv function including a high-level description, a list of arguments and their meanings, and more.
+The documentation for the `read_csv` function including a high-level description, a list of arguments and their meanings, and more.
 ```
 
 +++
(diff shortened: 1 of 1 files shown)


trevorcampbell (commented):
@joelostblom It looks like `help` works in python in general, while `?` only works in jupyter notebook / lab. It's also less "magicky" / keeps to standard Python syntax. I'm tempted to keep `help`. Two Qs for you:
- in the R version of the book, does our "help" method (I think we used `?` there?) work in the REPL as well, or just in jupyter?
  - `?` also works in the REPL. more reason to keep `help` in python
- is there any other advantage to `?` ?

source/intro.md
```diff
@@ -1057,18 +1057,17 @@ Bar plot of the ten Aboriginal languages most often reported by Canadian residen
...
 There are many Python functions in the `pandas` package (and beyond!), and
 nobody can be expected to remember what every one of them does
-or all of the arguments we have to give them. Fortunately, Python provides
-the `help` function, which
+or all of the arguments we have to give them. Fortunately, Jupyter provides
 provides an easy way to pull up the documentation for
-most functions quickly. To use the `help` function to access the documentation, you
-just put the name of the function you are curious about as an argument inside the `help` function.
+most functions quickly by using the `?` character. To access the documentation this way, you
+prefix `?` to the name of the function you are curious about.
```
trevorcampbell, line 1063: Get rid of these changes here -- keep `help` first. Then below where we say "if you're working in a Jupyter environment", you can introduce `?` as a convenience method that prints the help more nicely.

trevorcampbell: @joelostblom tagging in case commenting on your PR doesn't do that already :-)

trevorcampbell:
@joelostblom Ah, I just saw your comment on the other PR. I see `?` formats the help more nicely.

Let's just introduce both. We have to remember that this book is intended for a general audience, not just students of our class. So if someone reads this and tries `?` in the REPL, they might be stumped. 

See my comment above -- I think the best way is to keep the original `help` text, and since we already have a "if you're working in JLab" section below, just introduce `?` briefly there.

source/intro.md
```diff
@@ -1117,7 +1116,7 @@ The suggestions that are shown after typing `pd.read` and pressing <kbd>Tab</kbd
 To get more info on the function you want to use,
 you can type out the full name
 and then hold <kbd>Shift</kbd> while pressing <kbd>Tab</kbd>
-to bring up a help dialogue including the same information as when using `help()`.
+to bring up a help dialogue including the same information as when using `?`.
```
joelostblom, line 1120:
```suggestion
If you are working outside modern programming environments such as JupyterLab, VScode, and IPython,
you can access the help using the the `help()` function instead of `?`.

```

@trevorcampbell something like this? I don't like namedropping to new things to explain `help`, but I also don't want to get into what a REPL or "the basic python interpreter is".

joelostblom:
> in the R version of the book, does our "help" method (I think we used ? there?) work in the REPL as well, or just in jupyter?

  The help in R (which is `?`) works everywhere

> if someone reads this and tries ? in the REPL, they might be stumped.

In my mind our book is targeted towards novices in modern environments such as JupyterLab and VSCode. For example, if a reader tried creating a chart in the REPL they would be even more stumped and we don't offer any special section for explaining that. Another reason I avoided `help` was that I didn't want to get into explaining different environments too much, but I added a short sentence now.

Also note that some of our later worksheets (e.g. ch2) seem to already use `?` instead of `help`, just that it was never explained why.

source/intro.md
trevorcampbell: @joelostblom I was suggesting the other way around -- keep the original text on `help`, and then later in the JLab section introduce `?` as an alternative w/o much extra explanation. I agree we should avoid referencing technical stuff (REPLs etc) there, so we can just say it's specific to JLab and is nice because less typing.

I'll do a separate PR to show what I mean

trevorcampbell:
@joelostblom see #192 -- I prefer that edit. If you are OK with it, we can close this one and merge that one.

PR #188, opened 2023-08-01
Introduce `assign` in the first chapter

joelostblom:
Since it is used in the first worksheet and tutorial. https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/pull/794/files#r2524

### DIFF
--- a/source/intro.md
+++ b/source/intro.md
@@ -796,36 +796,6 @@ more easily understood format.
 Visualizations are a great tool for summarizing information to help you
 effectively communicate with your audience.
 
-## Adding new columns using `assign`
-
-What if we wanted to perform a computation on a column
-and then store it in the dataframe?
-For example,
-maybe we want to know what proportion of Canadian residents
-speak each of the most common Aboriginal languages as their mother tongue?
-To compute this quantity
-we need to divide the absolute number of residents with the total population of Canada.
-Note that we formatted the number for the Canadian population using underscores (`_`) so that it is easier to read;
-this does not affect how Python interprets the number and is just added for readability.
-
-```{code-cell} ipython3
-canadian_population = 35_151_728
-ten_lang['mother_tongue'] / canadian_population
-```
-
-To create a new column which stores the results of this operation,
-we can use the `assign` method:
-
-```{code-cell} ipython3
-ten_lang = ten_lang.assign(
-    mother_tongue_proportion=ten_lang['mother_tongue'] / canadian_population,
-)
-ten_lang
-```
-
-We can see that the 64,050 speakers of Cree
-correspond to only 0.18% of Canada's overall population.
-
 ## Exploring data with visualizations
 
 ```{index} visualization
--- a/source/intro.md
+++ b/source/intro.md
@@ -559,9 +559,7 @@ Syntax for using the `loc[]` operation to filter rows and select columns.
 ```
 
 ```{code-cell} ipython3
-aboriginal_lang = can_lang.loc[
-  can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]
-]
+aboriginal_lang = can_lang.loc[can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]]
 ```
 There is one very important thing to notice in this code example.
 The first is that we used the `loc[]` operation on the `can_lang` data frame by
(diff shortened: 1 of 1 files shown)


source/intro.md
```diff
@@ -245,7 +245,7 @@ is included with [the code for this book](https://github.com/UBC-DSCI/introducti
 If we were to open this data in a plain text editor (a program like Notepad that just shows
 text with no formatting), we would see each row on its own line, and each entry in the table separated by a comma:
 
-```code
+```text
```
trevorcampbell: Let's leave this as-is and do a new PR that fixes this throughout the book

we should also check that changing from `code` to `text` doesn't change formatting for the worse (and if it does, we should find the "right" language tag to use)

source/intro.md
```diff
@@ -760,6 +760,36 @@ more easily understood format.
 Visualizations are a great tool for summarizing information to help you
 effectively communicate with your audience.
 
+## Adding new columns using `assign`
+
```
trevorcampbell: this breaks the flow of the chapter. right before this the text is leading up to creating viz, and then it switches to assign

trevorcampbell (changes requested):
I'm not really sure introducing `assign` works well in this chapter. The intention for Ch1 was to get students all the way through a simple data analysis, without any unnecessary "branching paths." OTOH I think it would be nice to introduce assignment earlier than Ch3, where it is now.

I think if we want to introduce `assgin` we need to find a way to introduce it so that it is a "necessary step" along the path of the analysis. Along with that logic, it should also probably appear somewhere in section 1.7

Another general comment: we need to pair this with an equivalent PR in the R book.

source/intro.md
trevorcampbell: I'll open a PR for that now -- I'll just `sed` it

trevorcampbell: see #189

joelostblom:
I agree that the more our example is part of the critical path the better. Having that said, I think 1.7.4 (sorting) and 1.8 (chaining) already break the critical path of this chapter somewhat since neither is necessary wrangling to get to the viz. 1.7.4 might add some narrative, but 1.8 is to me a similar detour as the suggested new section with `assign`.

Again, I agree it would be better with a more apt example, but I'm running a bit short of time and can't see anything obvious to introduce, so I think I will leave this for now. I really think we should introduce `assign` in some way since it is so heavily used in the first two worksheets and tutorials, but we can give this some time and see if better ideas come up.

trevorcampbell:
> I really think we should introduce assign in some way since it is so heavily used in the first two worksheets and tutorials

I agree -- it was something that stuck out to me ages ago when we were first polishing the R material.

I'll take a stab at editing it on this PR and see what I come up with.

source/intro.md
```diff
@@ -559,9 +559,7 @@ Syntax for using the `loc[]` operation to filter rows and select columns.
 ```
 
 ```{code-cell} ipython3
-aboriginal_lang = can_lang.loc[
-  can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]
-]
+aboriginal_lang = can_lang.loc[can_lang["category"] == "Aboriginal languages", ["language", "mother_tongue"]]
```
trevorcampbell, line 562: We haven't covered multiline expressions yet, so put this back to the long single-line format

source/intro.md
```diff
@@ -586,7 +584,7 @@ with multiple kinds of `category`. The data frame
 `aboriginal_lang` contains only 67 rows, and looks like it only contains Aboriginal languages.
 So it looks like the `loc[]` operation gave us the result we wanted!
 
-### Using `sort_values` and `head` to select rows by ordered values
+## Using `sort_values` and `head` to select rows by ordered values
```
trevorcampbell, line 587: This didn't fit in the "using loc and []" subsection, so I moved it out to its own section

source/intro.md
```diff
@@ -631,6 +629,44 @@ ten_lang = arranged_lang.head(10)
 ten_lang
 ```
 
+## Adding and modifying columns using `assign`
+
```
trevorcampbell, line 633: I moved this earlier (to avoid breaking the lead-up to viz) and made it clear that this was a bit of a diversion (and a teachable moment for forking exploration ;) )

trevorcampbell (approved):
@joelostblom I made a few edits (see diff):
- reverted the `code` -> `text` syntax tag change (to avoid merge conflict with the other PR -- it was probably fine, but just in case)
- removed a multiline expression that was before we introduce those
- moved `sort` section out a level (didn't fit in the other subsection title)
- moved the new `assign` text earlier to avoid breaking up the flow with viz
- made it clear the new sub-question in the `assign` section is an exploration fork
  - I tried to write the chapter by changing the original data analysis question to ask for percentage/proportion, but it would spawn too many transitive-closure changes throughout that it was just too much work to do that way

If you are happy with it, feel free to merge

joelostblom:
LGTM, thank you!

PR #192, opened 2023-08-02
Adding `?` to Ch1 but keeping `help` first

trevorcampbell:
This is an alternative to #187 that keeps the more general `help` function first. 

(Also keeps the flow as-is, avoiding introducing Jupyter-only `?` and then later having a Jupyter-only section only to backtrack and introduce `help` later on.)


joelostblom (approved):
Looks good, although I still have a small preferences for focusing on `?`, I don't think it makes too big of a difference, so let's merge this and keep going.

PR #193, opened 2023-08-02
Use multi-line expressions more effectively

joelostblom:
I think it would be nice to put the parameters on their own lines when showing multiline expressions. I also changed the indent to 4 to match the rest of the book


trevorcampbell:
Thanks @joelostblom -- LGTM, will merge

PR #196, opened 2023-08-06
Remove fill encoding

joelostblom:
We don't teach `fill`


trevorcampbell (approved):
looks fine -- I have an issue out on the R book to go through and update learning objectives after this big round of editing is done. Will do Py book globally then too

PR #199, opened 2023-08-06
Sync clustering chapter with R

### LINKED ISSUE #176: Import data naming changes from tidymodels clustering edit R book
I did some renaming of data files in the clustering chapter as part of the tidyclust update. Should sync those here at some point.

trevorcampbell:
Matches the changes from https://github.com/UBC-DSCI/introduction-to-datascience/pull/496/files

Closes #176


source/clustering.md
```diff
@@ -66,22 +66,23 @@ exploratory analysis, i.e., uncovering patterns in the data.
...
+approach has both advantages and disadvantages.  Clustering requires no
+additional annotation or input on the data.  For example, while it would be
+nearly impossible to annotate all the articles on Wikipedia with human-made
+topic labels, we can cluster the articles without this information to find
+groupings corresponding to topics automatically.  However, given that there is
+no response variable, it is not as easy to evaluate the "quality" of a
+clustering.  With classification, we can use a test data set to assess
+prediction performance. In clustering, there is not a single good choice for
+evaluation. In this book, we will use visualization to ascertain the quality of
+a clustering, and leave rigorous evaluation for more advanced courses.
```
trevorcampbell, line 85: this paragraph was edited during our polishing pass on the R version, but the Py text was pulled from before that edit. So synced this to R.

source/clustering.md
```diff
@@ -174,7 +177,7 @@ from myst_nb import glue
 import pandas as pd
 
 data = pd.read_csv(
-    'data/toy_penguins.csv'
+    'data/penguins_toy.csv'
```
trevorcampbell, line 180: i'm trying to change the naming convention throughout to `[dataset-slug]_style.csv` to make the datasets sort nicely by filename. i'm just doing this as i go through, not a full pass on the whole repo

trevorcampbell: (similar change happened in R)

source/clustering.md
```diff
@@ -436,12 +439,14 @@ in the fourth iteration; both the centers and labels will remain the same from t
...
-> (2) the total WSSD is always greater than or equal to 0, and (3) there are only a finite number of possible
-> ways to assign the data to clusters. So at some point, the total WSSD must stop decreasing, which means none of the assignments
-> are changing, and the algorithm terminates.
+```{note}
+Is K-means *guaranteed* to stop at some point, or could it iterate forever? As it turns out,
+thankfully, the answer is that K-means is guaranteed to stop after *some* number of iterations. For the interested reader, the
+logic for this has three steps: (1) both the label update and the center update decrease total WSSD in each iteration,
+(2) the total WSSD is always greater than or equal to 0, and (3) there are only a finite number of possible
+ways to assign the data to clusters. So at some point, the total WSSD must stop decreasing, which means none of the assignments
+are changing, and the algorithm terminates.
```
joelostblom, line 448:
```suggestion
are changing, and the algorithm terminates. In practice, many implementations of KMeans
(including the one in scikit-learn) stops after a pre-set number of iterations 
when the probability of meaningful changes to the cluster labels is rather low.
```

Helpful or an unnecessary amount of details?

joelostblom (approved):
Looks good, just one minor comment

source/clustering.md
trevorcampbell: Yeah, I think this is too much detail for this book ;)

Issue #146, opened 2023-01-26
Be consistent in how we teach scikit-learn column transformers


joelostblom:
In some chapters we use the following and in some we don't:

```
from sklearn import set_config
set_config(transform_output="pandas")

make_column_transformer(
    ...
    verbose_feature_names_out=False,
)
```

It is new in 1.2 and seems useful as it removes the need to create a dataframe and keep track of column names manually.

joelostblom:
@trevorcampbell We currently set this option in ch 5 and 9. In ch 6-8, we don't convert the output to dataframes, so we don't need to set it which is why I think it is missing there. It is a bit inconvenient to have to write `    verbose_feature_names_out=False,` each time we create a column transformer. The alternative is to create the dataframe manually and keep track of the column names via `.get_feature_names_out()`, which is also inconvenient.

What do you think we should do here for the worksheets? Worksheet 5 currently does not use `set_config`. I can standardize when I go through them as long as we agree to something here first.

Ch5:
![image](https://github.com/UBC-DSCI/introduction-to-datascience-python/assets/4560057/ae1cfb6d-7f78-4d91-85bd-5c53fb7ae1b1)

trevorcampbell:
@joelostblom if I understand correctly, the most natural thing is to use `set_config(transform_output="pandas")` in the textbook and worksheets. Would that cause any problems in the textbook?

Basically I'd go with whatever creates the simplest code everywhere here. Thoughts?

joelostblom:
I incorrectly thought that we only needed `verbose_feature_names_out=False` if we used  `set_config(transform_output="pandas")`, but it turns out that they are also needed if we use `.get_feature_names_out()`. So let's go with the pandas option, I will close this and make the changes in the assignments. I will also add this to each textbook chapter so that students learn to always do it when working with sklearn.

There is discussion here to allow specifying a global option for `verbose_feature_names_out` which would make this more convenient https://github.com/scikit-learn/scikit-learn/issues/24724

PR #200, opened 2023-08-07
Always set the transformed output to pandas

joelostblom:
I think it will make life easier for students if we teach them to always set this option when using sklearn. Follow up on https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/146


trevorcampbell (approved):
LGTM

PR #201, opened 2023-08-17
Remove redundantly extracting the best estimator before predicting

joelostblom:
This is not needed and I don't think that useful. I made the same change in the worksheet, where we also ask them to extract the best params so I switched to that here.

### DIFF
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -807,15 +807,22 @@ To assess how well our model might do at predicting on unseen data, we will
 assess its RMSPE on the test data. To do this, we first need to retrain the 
 KNN regression model on the entire training data set using $K =$ {glue:}`best_k_sacr`
 neighbors. Fortunately we do not have to do this ourselves manually; `scikit-learn`
-does it for us. We just need to obtain the `best_estimator_` attribute of original
+does it for us automatically. If we want to see which parameters were identified
+as the best ones during grid search, we can access the `best_params_` attribute of the original
 fit `GridSearchCV` object.
 
 ```{code-cell} ipython3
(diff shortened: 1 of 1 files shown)


source/regression1.md
```diff
@@ -807,15 +807,22 @@ To assess how well our model might do at predicting on unseen data, we will
 assess its RMSPE on the test data. To do this, we first need to retrain the 
 KNN regression model on the entire training data set using $K =$ {glue:}`best_k_sacr`
 neighbors. Fortunately we do not have to do this ourselves manually; `scikit-learn`
-does it for us. We just need to obtain the `best_estimator_` attribute of original
+does it for us automatically. If we want to see which parameters were identified
```
trevorcampbell: I agree with this change, but this particular way of editing it breaks the flow -- we previously had this here because we said
- you now need to retrain with best param
- scikit does it for you
- here is the object to use below that has the best params

Now it's a bit weird:
- you now need to retrain
- scikit does it for you
- here is the best param value (but we already knew that from before)
- moving on without reference to this again

But I agree that just calling `.predict` on the fit object is far more natural, and we should just switch to that.

I'm going to move the `best_params_` up to right after we plot the CV results -- there it makes sense since we can show people how to grab the best result without visually inspecting the plot

trevorcampbell: I think we should also add something similar to classification 2 -- i can handle that

trevorcampbell: (with corresponding change to R version)

trevorcampbell:
OK, looks good now -- merging

PR #202, opened 2023-08-17
Remove/replace loc[:, cols], implicit data frame conversion, using functions instead of methods, and inconsistent spacing

joelostblom:
These are a few fixes that are not necessarily related but touch the same bit of code so it would lead to conflicts to make them as separate PRs. Happy to roll back individual commits if there is something that is not suitable.

- Remove use of `loc[:, list_of_cols]`
- Improve code formatting
- Prefer data frame methods over global functions
- Explain why we are converting to a data frame and be explicit and consistent when we do it

### DIFF
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -611,7 +611,7 @@ and rename the parameter column to be more readable.
 ```{code-cell} ipython3
 # fit the GridSearchCV object
 sacr_fit = sacr_gridsearch.fit(
-    sacramento_train[["sqft"]],  # A single column data frame
+    sacramento_train[["sqft"]],  # A single-column data frame
     sacramento_train["price"]  # A series
 )
 
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -306,7 +306,7 @@ of a house that is 2,000 square feet.
 
 ```{code-cell} ipython3
 nearest_neighbors = (
-    small_sacramento.assign(diff=abs(2000 - small_sacramento["sqft"]))
+    small_sacramento.assign(diff=(2000 - small_sacramento["sqft"]).abs())
     .nsmallest(5, "diff")
 )
 
@@ -1018,7 +1018,7 @@ sacr_results["mean_test_score"] = -sacr_results["mean_test_score"]
 
 # show only the row of minimum RMSPE
 sacr_results[
-   sacr_results["mean_test_score"] == min(sacr_results["mean_test_score"])
+   sacr_results["mean_test_score"] == sacr_results["mean_test_score"].min()
 ]
 ```
 
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -186,7 +186,7 @@ want to predict (sale price) on the y-axis.
 ```{code-cell} ipython3
 :tags: [remove-output]
 
-eda = alt.Chart(sacramento).mark_circle().encode(
+scatter = alt.Chart(sacramento).mark_circle().encode(
     x=alt.X("sqft")
         .scale(zero=False)
         .title("House size (square feet)"),
@@ -195,12 +195,12 @@ eda = alt.Chart(sacramento).mark_circle().encode(
         .title("Price (USD)")
 )
 
-eda
+scatter
 ```
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
-glue("fig:07-edaRegr", eda)
+glue("fig:07-edaRegr", scatter)
 ```
 
 :::{glue:figure} fig:07-edaRegr
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -600,19 +600,20 @@ and rename the parameter column to be more readable.
 ```{code-cell} ipython3
 # fit the GridSearchCV object 
 sacr_fit = sacr_gridsearch.fit(
-                  sacramento_train[["sqft"]],
-                  sacramento_train[["price"]]
-              )
-# retrieve the CV scores
-sacr_results = pd.DataFrame(sacr_fit.cv_results_)[
-    ["param_kneighborsregressor__n_neighbors", "mean_test_score", "std_test_score"]
-]
-sacr_results = sacr_results.assign(
-    sem_test_score = sacr_results["std_test_score"] / 5**(1/2)
-).rename(
-    columns = {"param_kneighborsregressor__n_neighbors" : "n_neighbors"}
-).drop(
-    columns = ["std_test_score"]
+    sacramento_train[["sqft"]],
+    sacramento_train[["price"]]
+)
+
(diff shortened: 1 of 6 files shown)


source/regression1.md
```diff
@@ -306,7 +306,7 @@ of a house that is 2,000 square feet.
 
 ```{code-cell} ipython3
 nearest_neighbors = (
-    small_sacramento.assign(diff=abs(2000 - small_sacramento["sqft"]))
+    small_sacramento.assign(diff=(2000 - small_sacramento["sqft"]).abs())
```
joelostblom, line 309: I made this change in the worksheet too

source/regression1.md
```diff
@@ -596,23 +596,28 @@ on `sacr_gridsearch`. As we did in the {ref}`classification2` chapter,
...
-sacr_results = pd.DataFrame(sacr_fit.cv_results_)[
-    ["param_kneighborsregressor__n_neighbors", "mean_test_score", "std_test_score"]
-]
-sacr_results = sacr_results.assign(
-    sem_test_score = sacr_results["std_test_score"] / 5**(1/2)
-).rename(
-    columns = {"param_kneighborsregressor__n_neighbors" : "n_neighbors"}
-).drop(
-    columns = ["std_test_score"]
+    sacramento_train["sqft"].to_frame(),
```
joelostblom: I think it is important to be explicit here. Previously we just used double bracket `[[...]]` with a single column name, which does return a dataframe, but it is not something we discuss anywhere in the book. We could add a discussion of this also and keep using double bracket for a single column name but `to_frame()` is a bit more versatile since it can be used in chains too and it might be confusing that we sometimes use single and sometimes double bracket with a single column

trevorcampbell: @joelostblom rather than introducing a new function, why not just remind readers that `[...]` outputs a DF with a list of colns, and `["blah"]` is a list with a single element, so `[["blah"]]` outputs a df with one coln.

I think I might prefer that over introducing a new function here when there's a perfectly good tool the students already know

source/regression2.md
```diff
@@ -399,8 +399,8 @@ intercept of the line via the `intercept_` property.
 # fit the linear regression model
 lm = LinearRegression()
 lm.fit(
-   sacramento_train[["sqft"]],
-   sacramento_train[["price"]]
+   sacramento_train["sqft"].to_frame(),
```
trevorcampbell: same comment here

source/regression2.md
```diff
@@ -446,7 +446,7 @@ we predict on the test data set to assess how well our model does.
 ```{code-cell} ipython3
 # make predictions
 sacr_preds = sacramento_test.assign(
-    predicted = lm.predict(sacramento_test[["sqft"]])
+    predicted = lm.predict(sacramento_test["sqft"].to_frame())
```
trevorcampbell: same comment here

trevorcampbell:
@joelostblom this PR looks great -- I think we should stick with `[[...]]` instead of introducing `to_frame`. If you swap that back we can merge, or if you feel strongly we can discuss more.

source/regression1.md
joelostblom: I agree that we could change and explain the difference between extracting a single column with single or double brackets. The reason that I went with `to_frame` is because there is [a previous  worksheet](https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/pull/797/files#r2574) where we need to use `to_frame` in  a method chain (where we can't use a single element list), so we need to explain that somewhere. If we go with the double brackets solution here, we could wrap the method chain in the worksheet in a call to  `pd.Dataframe` instead; it looks slightly less neat but not a huge deal.

After completing some of the machine learning worksheets, one more advantage of using `pd.Dataframe` over `to_frame` is that there are some dictionaries from `cv_results_` that we wrap in `pd.Dataframe` so using the same method for converting series to dataframes might make it easier to remember and they will not mix up when they can use `to_frame` and when they can't.

So all in all, I think I'm also leaning towards `pd.DataFrame` over `to_frame`. What do you think about preferring `pd.DataFrame` even over double brackets for a single column? So for the code above it would be `pd.DataFrame(sacramento_train["sqft"])`. Then things are very explicit and we always use the same method to convert a series to a dataframe, but the code is a bit inelegant and more tedious to write.

trevorcampbell: The constructor call is definitely too bulky to be used everywhere. I think the best option is to go with the simplest one in each scenario (and properly introduce them the first time they're used, or a reminder/pointer if the introduction is far enough away and they aren't used regularly). My order of preference is:
1. `[[...]]`
2. `(...).to_frame()`
3. `pd.DataFrame(...)`

`[[....]]` handles basically every case we encounter. As long as we properly introduce this the first time it's done, it should be fine (since it will probably be seen somewhat frequently throughout). Then `to_frame` can get used where it's strictly necessary, with perhaps a reminder for the reader why we use it since it'll be somewhat infrequent. `pd.DataFrame` I imagine only is really needed for converting dictionaries, so again we can mention that there the one (?) time it happens.

joelostblom: Sounds good! I just updated this PR with the changes, including a mention of single column dataframes in the wrangling chapters, have a look. I will go through the assignments now; if there is only a single mention of `to_frame()` throughout all assignment I might remove it altogether for simplicity.

source/regression1.md
```diff
@@ -596,23 +596,31 @@ on `sacr_gridsearch`. As we did in the {ref}`classification2` chapter,
...
+)
+
+# Retrieve the CV scores
+sacr_results = (
+    pd.DataFrame(sacr_fit.cv_results_)[[
+        "param_kneighborsregressor__n_neighbors",
+        "mean_test_score",
+        "std_test_score"
+    ]]
+    .assign(sem_test_score=sacr_results["std_test_score"] / 5**(1/2))
```
trevorcampbell, line 626: @joelostblom this generates an error in the textbook. `sacr_results` hasn't been created yet -- this is why it was done in two steps before. I will fix it.

source/regression1.md
```diff
@@ -999,25 +1010,25 @@ sacr_fit = GridSearchCV(
...
-).drop(
-    columns = ["std_test_score"]
-)
+sacr_results = (
+    pd.DataFrame(sacr_fit.cv_results_)[[
+        "param_kneighborsregressor__n_neighbors",
+        "mean_test_score",
+        "std_test_score"
+    ]]
+    .assign(sem_test_score=sacr_results["std_test_score"] / 5**(1/2))
```
trevorcampbell, line 1030: @joelostblom this is incorrect for the same reason as above, but this time it will fail silently (it'll use the old `sacr_results` df from before)

I will fix it

PR #203, opened 2023-08-17
Remove categorical type casting

### LINKED ISSUE #197: Remove or expand on pandas categoricals
Currently this is all we mention about pandas categoricals in the entire book:

image

Everywhere else we use categorical values as strings/objects. I think it is confusing to mention this only at one place and suggest that we remove it until we are ready to write a longer section on it.

joelostblom:
close #197 

Until we have time to expand this section, I think we should remove it to not confuse students and suddenly start using the categorical data type after not doing so for the first four chapters.


trevorcampbell:
LGTM (after a minor wording edit) -- thanks @joelostblom -- merged

PR #207, opened 2023-08-19
Remove transform regression in favor of using the same sklearn predict approach everywhere

joelostblom:
I think we should avoid teaching `transform_regression` for linear regression in Altair in favor of using the same approach we used for KNN regression (predicting with scikit-learn). Although `transform_regression` is a convenience when plotting data, I don't think there is much gained from introducing this when we are already using sklearn to build a model; it shines when not building the model in the first place. Having a special way of doing things for linear regressions and another way of doing things for all other models could be a bit confusing and it is the only "transform" method we teach (without explaining what they are) which could also be confusing.

I made these changes in the worksheet too here https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/pull/815. If we like this change we can consider doing it in the R textbook and worksheet too. I think that could be in a separate PR about geom_smooth.


source/regression1.md
```diff
@@ -862,20 +862,31 @@ could make or break whether or not they could afford put an offer on a house.
...
+})
+# Predict the price for each of the sqft values in the grid
+sacr_preds = sqft_prediction_grid.assign(
+    predicted = sacr_fit.predict(sqft_prediction_grid)
 )
 
-# the base plot: the training data scatter plot
-base_plot = alt.Chart(sacramento_train).mark_circle().encode(
+# Plot all the houses
+base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
```
joelostblom, line 905: I changed to all the data here (and in the next chapter) instead of just the training data to match the text. However, we could consider bigger changes to this snippet and creating a prediction for each of the values in the test dataset and then plot only the test data. This is what we do in the worksheets, and it would fit thematically here in the textbook since this section is under the heading "Evaluating on the test set". I am not sure about the initial motivation for comparing against all the data here?

trevorcampbell:
@joelostblom I totally agree with the change to get rid of the special plotting regression stuff (this has always bugged me a bit, but I never got around to fixing it) and replacing it with predictions from the models that were already built.

Can you please make the same changes in the R version (with appropriate PRs etc) so I can review all at once?

joelostblom:
Done. The textbook changes are here https://github.com/UBC-DSCI/introduction-to-datascience/pull/520 and the worksheet changes are here https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/pull/815

source/regression1.md
joelostblom: @trevorcampbell  What do you think of this comment? Currently the textbook only shows how to create predictions using the manual grid and then compare it with the all the data. In the assignments, we ask students to create predictions from the test data and plot it over the test data only.

trevorcampbell: Let's just open an issue for that. I'm reluctant to mess with this right now too much -- when we originally wrote the book I remember this being a notable point of discussion, and in order to properly address this I'd need to go back and refresh my memory on all the various arguments surrounding.

PR #210, opened 2023-08-21
Remove section on how to use assign to create new data frames

joelostblom:
I'm unsure about the value in using `assign` for creating new dataframes. It is not something I have seen before myself and I don't see a corresponding section in the R book either. I suggest removing it in favor of being explicit about that dictionaries is the most convenient way to create new data frames, particularly since this is what we use ourselves everywhere else.


trevorcampbell:
I polished this up, and kept the section on `merge` because I think it's still useful (but I decoupled it from `assign` and used it as another example of creating dfs from `dict`s)

PR #211, opened 2023-08-22
Minor clarifications to clustering chapter

joelostblom:


trevorcampbell (approved):
@joelostblom I had to make transitive closure edits on the R version -- please make sure to open a PR there any time you make changes that should be propagated.

Also: The R version now uses a `workflow` for tuning the number of neighbours. I think we wrote the py version previously to mimic the old method (manually tuning). Is there a more convenient way of tuning the number of neighbours in `scikit-learn`?

I will open an issue for that discussion, and merge this one.

Issue #80, opened 2022-12-31
Removing deprecation warnings


trevorcampbell:
Some `altair` code spits out a deprecation warning currently.

To suppress these, I'm currently including the code

```
{code-cell} ipython3
:tags: [remove-cell]
import warnings
def warn(*args, **kwargs):
    pass
warnings.warn = warn
```

to just eat all the warnings. But later once `altair` updates / book stabilizes, we should delete this code and make sure we aren't skipping over any important warnings.

joelostblom:
Two thoughts on this:

1. I think we can use the following to just suppress the deprecation warning and not risk missing something important:
    ```py
    import warnings
    warnings.filterwarnings("ignore", category=DeprecationWarning)
    ```
2. Regardless of how we suppress the warnings in the book, students will still see them when they work with JupyterLab, so it could still be a good idea to restrict pandas to <1.5 on the server to avoid students getting confused by always seeing that warning when they use altair.

trevorcampbell:
> I think we can use the following to just suppress the deprecation warning and not risk missing something important:

This is actually what I tried originally, but it didn't work. I will try it again...

> Regardless of how we suppress the warnings in the book, students will still see them when they work with JupyterLab

This is not an issue. I'm more concerned about the polish on the book -- if anyone is scared of these warnings in class we can just explain it to them.

trevorcampbell:
needed

```
warnings.filterwarnings("ignore", category=FutureWarning)
```

to actually prevent the altair warnings

joelostblom:
This has been added to all the relevant chapters so closing the issue

PR #218, opened 2023-08-29
Fix PDF build and reduce chart sizes in HTML build

### LINKED ISSUE #143: Reduce chapter size by removing chart data
We should explore using either Altair saver's svg renderer with the data server, or look into VegaFusion which are both options to not include all the data in the charts, which currently make some of our lecture pages huge:

image

### LINKED ISSUE #171: Fix render of PDF chart images
@joelostblom I've got the PDF build working now (just run `build_pdf.sh`), but all the charts don't render properly. They all look like:

image

Perhaps you know the fix for this? I tried playing with stuff [here](https://altair-viz.github.io/user_guide/display_frontends.html) but no luck.

joelostblom:
This fixes the PDF issues for me. Note that you need to install `altair 5.1` (note that there is a minor hotfix coming in 5.1.1 this week as well, not superimportant) and `vl-convert-python `0.13 on the server. For the HTML version, we can use `vegafusion[embed]` (also needs an install) to reduce the chart sizes somewhat; it avoids storing the entire dataframe and only uses the data points needed for a chart which is a great difference for e.g. a histogram, but not as much for a scatter plot). This also raises the limit of how big datasets can be plotted and in many cases is a replacement for `disable_max_rows()` (probably in all cases for us). I can add this to the assignments as well if you think that is a good idea (instead of disable_max rows)

Before:
image

After:
Screenshot from 2023-08-28 22-17-47

close https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/171
close https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/143


source/classification1.md
```diff
@@ -388,13 +398,10 @@ Scatter plot of concavity versus perimeter with new observation represented as a
...
-    (
-        cancer.loc[np.argmin(my_distances), attrs],
-        perim_concav_with_new_point_df.loc[len(cancer), attrs],
-    ),
-    axis=1,
-).T
+near_neighbor_df = pd.concat([
+    cancer.loc[[np.argmin(my_distances)], attrs],
+    perim_concav_with_new_point_df.loc[[cancer.shape[0]], attrs],
+])
```
joelostblom, line 390: Due to a small regression in altair 5.1 (to be fixed in 5.1.1) "object" columns are not longer automatically converted to quantitative types. We had some odd concat code that created objects where ints should be which I fixed, plus added some explicit types

source/classification1.md
```diff
@@ -22,6 +22,16 @@ from myst_nb import glue
 import numpy as np
 from sklearn.metrics.pairwise import euclidean_distances
 from IPython.display import HTML
+import altair as alt
+import os
+
+
+# Use PNG images in the PDF version of the books to make sure that they render
+if os.environ['BOOK_BUILD_TYPE'] == 'PDF':
```
trevorcampbell: this will generate a key error when building in html -- you need to `export BOOK_BUILD_TYPE='HTML'` in the build html script

source/classification1.md
trevorcampbell: @joelostblom is this diff trying to match the temporarily broken altair 5.1, or the new fixed 5.1.1? Assuming 5.1.1 is coming soon, I'd prefer to just wait to merge this until 5.1.1 and follow that syntax.

trevorcampbell: perhaps I am not understanding your comment well...is this new code here related to the regression in altair?

source/classification2.md
```diff
@@ -1242,7 +1250,7 @@ to layer a point and line chart.
 :tags: [remove-output]
 
 accuracy_vs_k = alt.Chart(accuracies_grid).mark_line(point=True).encode(
-    x=alt.X("n_neighbors").title("Neighbors"),
+    x=alt.X("n_neighbors:Q").title("Neighbors"),
```
trevorcampbell: maybe your comment above is referring to this.... sorry, I'm a bit lost

trevorcampbell (changes requested):
Thanks @joelostblom . I need some clarification on one of your comments. Also two additional things are needed:
- need to edit the `build_html.sh` script to edit the environment
- need to edit the Dockerfile to add any missing packages you mentioned in your message

trevorcampbell:
@joelostblom FYI: once you resolve the review comments, I'm planning to merge this and push out the new production version for this semester

source/_config.yml
```diff
@@ -78,6 +78,8 @@ sphinx:
   local_extensions: # A list of local extensions to load by sphinx specified by "name: path" items
   config: # key-value pairs to directly over-ride the Sphinx configuration
     bibtex_reference_style: author_year
+  html_context:
+    default_mode: light
```
joelostblom, line 82: Default to light background for the book, instead of system default.

source/chapter_preamble.py
joelostblom, line 1: I moved all the preamble/setup code into a single file so that we don't need to modify this in multiple places if we tweak e.g. the dpi or figure size. I also moved the common imports here (`glue` and `os`) for convenience

source/classification1.md
```diff
@@ -14,14 +14,10 @@ kernelspec:
...
 ```{code-cell} ipython3
 :tags: [remove-cell]
-import warnings
-warnings.filterwarnings("ignore", category=DeprecationWarning)
-warnings.filterwarnings("ignore", category=FutureWarning)
 
-from myst_nb import glue
-import numpy as np
-from sklearn.metrics.pairwise import euclidean_distances
+from chapter_preamble import *
```
joelostblom, line 18: This runs the setup code and imports the imports from the chapter preamble here, which is why we no longer need the explicit imports of glue and os in each notebook

source/classification1.md
```diff
@@ -14,14 +14,10 @@ kernelspec:
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
-import warnings
-warnings.filterwarnings("ignore", category=DeprecationWarning)
-warnings.filterwarnings("ignore", category=FutureWarning)
```
joelostblom, line 19: These warnings no longer arise since we updated altair so I have removed the silencing so that we can spot new warnings as mentioned in https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/80. The only current ones I found are the ones for the default n_init for kmeans in scikit-learn, which will go away in 1.4; we could silence them now if we think they are ugly but I'll leave it to you to decide

source/classification1.md
joelostblom: Yes, this is the new code related to the regression in altair, but it is also a better way of doing the concatenation since it is less complex and result in the correct data types being propagated instead of numbers beings cast as objects

source/classification2.md
joelostblom: This can now be removed since we use 5.1.1 already. The reason for doing this is that the parameters from a grid search are returned as objects instead of ints/floats (not sure why). Now that altair again uses `infer_objects()` as long as the input is a pd dataframe, these objects are cast as ints internally and we don't need to do it explicitly.

source/classification2.md
```diff
@@ -292,6 +284,7 @@ on the series of numbers, passing the argument `n = 10` to indicate that we want
 
 ```{code-cell} ipython3
 import numpy as np
+import pandas as pd
```
joelostblom, line 287: Pandas was used visibly but the import was hidden

source/clustering.md
```diff
@@ -632,6 +630,10 @@ visualize them as shown in {numref}`cluster_plot`.
 Note that we are plotting the *un-standardized* data here; if we for some reason wanted to 
 visualize the *standardized* data, we would need to use the `fit` and `transform` functions
 on the `StandardScaler` preprocessor directly to obtain that first.
+As in Chapter {ref}`viz`,
```
joelostblom, line 633: I added a reminder here so that students remember since the viz chapter was a while ago

source/inference.md
```diff
@@ -1043,7 +1037,7 @@ alt.Chart(six_bootstrap_samples, height=150).mark_bar().encode(
         .title("Price per night (dollars)"),
     y=alt.Y("count()").title("Count")
 ).facet(
-    "replicate",
+    "replicate:N",  # Remeber that `:N` converts the variable to a categorical type
```
joelostblom: These two occurrences of a suffix inside `facet` are difference than the 5.1 regression. Details are in https://github.com/altair-viz/altair/issues/3138, but in brief altair suppresses some warnings from vega-lite while vegafusion does not. So after switching the data_transformer we would see the following warning without the suffix:

image

joelostblom (commented):
@trevorcampbell Ready for another look. I changed things around a bit so that it is more convenient to make changes in the preamble that we want propagated to all notebooks.

Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN rm -rf work
 RUN pip install docutils==0.17.1 # Need to pin docutils to an old version for now, due to https://github.com/executablebooks/jupyter-book/issues/2022
 RUN pip install referencing
 RUN pip install jupyter-book
-RUN pip install numpy jinja2 altair_data_server vl-convert-python click ibis-framework ghp-import jupytext nodejs
+# Pinning pandas until altair 5.1.2 to avoid future warning https://github.com/altair-viz/altair/issues/3181
+RUN pip install numpy jinja2 pandas"<2.1" altair">=5.1.1" "vegafusion[embed]" vl-convert-python">=0.13" click ibis-framework ghp-import jupytext nodejs
```
joelostblom, line 21: Although I added these here, it seems that I still need `pip install ...` in the build files. I thought the docker file would automatically update the container, but maybe there is a manual step involved to update the docker file?

trevorcampbell:
FYI @joelostblom I'm going to merge `main` back into this branch to get the new Dockerfile (which has `altair==5.1.1`)

Dockerfile
trevorcampbell: @joelostblom the dockerfile only rebuilds when commiting to `main` (otherwise we would have to do more gymnastics with temporary docker images etc. i'll open an issue about it, but for now I'll merge `main` into this branch to get the new updated dockerfile

trevorcampbell:
Actually after fixing a few minor issues, I'm just going to merge this and work directly on `main` to clean up last things before the push to `production`

PR #227, opened 2023-08-31
Fix vegafusion warnings

### LINKED ISSUE #223: Warnings in build (from vegalite?)
@joelostblom after merging that recent PR for PDF builds, I see some new warnings:

In inference:
```
/home/jovyan/source/inference.md: Executing notebook using local CWD [mystnb]  
WARN Dropping {} from channel "y" since it does not contain any data field, datum, value, or signal.
WARN Dropping {} from channel "y" since it does not contain any data field, datum, value, or signal.
WARN text dropped as it is incompatible with "rule".
WARN text dropped as it is incompatible with "rule".
```
In viz:
```
/home/jovyan/source/viz.md: Executing notebook using local CWD [mystnb]        
WARN facet encoding should be discrete (ordinal / nominal / binned).
WARN facet encoding should be discrete (ordinal / nominal / binned).
WARN facet encoding should be discrete (ordinal / nominal / binned).
WARN facet encoding should be discrete (ordinal / nominal / binned).
WARN facet encoding should be discrete (ordinal / nominal / binned).
WARN facet encoding should be discrete (ordinal / nominal / binned).
```

I fixed one instance in `classification2.md` on `main` already, but these two persist.

joelostblom:
close https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/223

### DIFF
--- a/source/inference.md
+++ b/source/inference.md
@@ -1251,18 +1251,19 @@ visualize the interval on our distribution in {numref}`fig:11-bootstrapping9`.
 
 ```{code-cell} ipython3
 # Create the annotation for for the 2.5th percentile
-text_025 = alt.Chart().mark_text(
+rule_025 = alt.Chart().mark_rule(color='#f58518', size=3, strokeDash=[5]).encode(
+    x=alt.datum(ci_bounds[0.025])
+).properties(
+    width=500
+)
+text_025 = rule_025.mark_text(
     color='#f58518',
     size=12,
     fontWeight='bold',
     dy=-160
 ).encode(
-    x=alt.datum(ci_bounds[0.025]),
     text=alt.datum(f'2.5th percentile ({ci_bounds[0.025].round(1)})')
-).properties(
-    width=500
 )
-rule_025 = text_025.mark_rule(color='#f58518', size=3, strokeDash=[5])
 
 # Create the annotation for for the 97.5th percentile
 text_975 = text_025.encode(
--- a/source/inference.md
+++ b/source/inference.md
@@ -1122,11 +1122,12 @@ the true sampling distribution&mdash;which corresponds to taking many samples fr
 ```{code-cell} ipython3
 :tags: [remove-input]
 
+sampling_distribution.encoding.x['bin']['extent'] = (90, 250)
 alt.vconcat(
     alt.layer(
         sampling_distribution,
-        sampling_distribution.mark_rule(color='#f58518', size=2).encode(x='mean(mean_price)', y=alt.Y()),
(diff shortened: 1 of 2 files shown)


source/inference.md
```diff
@@ -1122,27 +1122,26 @@ the true sampling distribution&mdash;which corresponds to taking many samples fr
...
+        alt.Chart(boot20000_means).mark_rule(color='#f58518', size=2).encode(x='mean(mean_price)'),
+        alt.Chart(boot20000_means).mark_text(color='#f58518', size=12, align='left', dx=18, fontWeight='bold').encode(
             x='mean(mean_price)',
-            y=alt.value(6),
+            y=alt.value(7),
             text=alt.value(f"Mean = {boot_est_dist['data']['mean_price'].mean().round(1)}")
         )
     ).properties(title='Bootstrap distribution', height=150)
-).resolve_scale(
-    x='shared'
```
joelostblom, line 1145: I changed this to use the same approach as elsewhere in the textbook with setting the extent instead of the scale for binned scales. This change also works around this bug https://github.com/hex-inc/vegafusion/issues/382

source/inference.md
```diff
@@ -1252,18 +1251,19 @@ visualize the interval on our distribution in {numref}`fig:11-bootstrapping9`.
 
 ```{code-cell} ipython3
 # Create the annotation for for the 2.5th percentile
-text_025 = alt.Chart().mark_text(
+rule_025 = alt.Chart().mark_rule(color='#f58518', size=3, strokeDash=[5]).encode(
+    x=alt.datum(ci_bounds[0.025])
+).properties(
+    width=500
+)
+text_025 = rule_025.mark_text(
```
joelostblom, line 1259: Avoid using text encoding with unsuppoerted mark type

source/viz.md
```diff
@@ -1554,13 +1554,14 @@ We also reduce the height of each chart
...
+We also explicitly specify that `facet` is a categorical variable
+since faceting should only be done with categorical variables.
 
 ```{code-cell} ipython3
-
 morley_hist_facet = morley_hist_categorical.properties(
     height=100
 ).facet(
-    "Expt",
+    "Expt:N",
```
joelostblom, line 1564: This is to avoid the warning. It is technically possible to use a non-categorical variable with facet, which is why I chose the word "should".

joelostblom (commented):
@trevorcampbell I think this fixes the build warnings; have a look.

trevorcampbell (approved):
LGTM, thanks @joelostblom

PR #237, opened 2023-09-11
Add back missing argument

joelostblom:
Not sure when this was removed, but it seems to be accidental since it no longer matches the text so adding it back


trevorcampbell:
LGTM. Thanks @joelostblom . This isn't a major problem though, so I'll hold off on re-release to avoid accidental problems mid-semester.

PR #241, opened 2023-09-13
Trigger Dockerfile rebuild/commit on PR

### LINKED ISSUE #221: Dockerfile temporary image build in PRs
Right now if we want to try out new edits to a dockerfile in a PR, we have to edit on `main` and merge back into the PR (potentially temporarily breaking `main`). Would be good to edit the workflow so that we can build temp docker images in PRs for testing.

trevorcampbell:
This PR would edit the github action to rebuild the docker image when a PR is opened targeting `Dockerfile`. That would let us easily review PRs that edit the Dockerfile without needing to manually build things locally or interactively before testing.

One TODO: I would need to enable `main` branch protection to stop accidental commits to `Dockerfile` directly, which would circumvent the action.

There is one problem with this approach: if someone else were to fork this repo and make a PR that edits the `Dockerfile`, the workflow would fail. That is very unlikely to happen, so maybe the benefit of being able to check the build on most PRs is worth it.

Another approach would be to have github action build a copy of the book using dockerhub, or if it can't find the image, build the image first locally w/o pushing to dockerhub. That would work with forks, and would enable PR review without needing to build the image locally. Big downside is needing to enable an app (netlify etc) to have permissions in our repos, which I'm not a fan of.

Closes #221 

(We should take the same approach in the R book repo)

@joelostblom thoughts?

### DIFF
--- a/.github/workflows/update_build_environment.yml
+++ b/.github/workflows/update_build_environment.yml
@@ -1,8 +1,6 @@
 name: Rebuild and publish new ubcdsci/py-intro-to-ds image on DockerHub
 on:
-  push:
-    branches:
-      - main
+  pull_request:
     paths:
       - Dockerfile
 jobs:
@@ -11,11 +9,11 @@ jobs:
     permissions:
       contents: write
     steps:
-    - name: Checkout main
+    - name: Checkout PR branch
       uses: actions/checkout@v3
       with:
         fetch-depth: '0'
-        ref: 'main'
+        ref: ${{ github.head_ref }}
     - name: Rebuild and publish image
       id: rebuild
       uses: elgohr/Publish-Docker-Github-Action@v5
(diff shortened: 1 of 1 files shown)


trevorcampbell:
The more I think about it, the more I'm happy with the way this branch does it. I don't think we've ever (either Py or R) had a fork make a PR to `Dockerfile`.

joelostblom:
I'm in favor of QoL improvements like this that facilitates reviewing! I'm not worried about this not triggering for forks; as you said we don't really use that workflow. I think this gets us closer to what we discussed a long time ago about automatic builds in https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/44, and the dockerhub approach you described would get us all the way there so I'm intrigued although I don't know how much additional work it would be (I don't mind the enabling of app permissions personally).

trevorcampbell:
Alright I'll mark this ready for review, and we can test it out in this repo a bit before doing the same in the R repo.

joelostblom (approved):
Looking at the diff, the changes here make sense to me. However, I have limited experience with actions workflow, so I wouldn't put much trust in my ability to detect if an undesired change was introduced here.

PR #242, opened 2023-09-13
Add note on plotting large datasets

### LINKED ISSUE #136: Talk about how to handle large data sets in Altair

joelostblom:
Close https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/136. Let's not put this into production before this years students are past the viz chapter, just to avoid confusion.

I opted to put the note early on, similar to how we did it in the database section where we mention sqlalchemy up front. It also fits withthat the transformer line would go right after the import.

### DIFF
--- a/source/viz.md
+++ b/source/viz.md
@@ -160,7 +160,7 @@ understand and remember your message quickly.
 +++
 
 ## Creating visualizations with `altair`
-#### Build the visualization iteratively
+#### *Build the visualization iteratively*
 
 ```{index} altair
 ```
--- a/source/viz.md
+++ b/source/viz.md
@@ -160,7 +160,7 @@ understand and remember your message quickly.
 +++
 
 ## Creating visualizations with `altair`
-#### *Build the visualization iteratively*
+#### Build the visualization iteratively
 
 ```{index} altair
 ```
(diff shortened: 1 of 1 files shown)


source/viz.md
```diff
@@ -160,7 +160,7 @@ understand and remember your message quickly.
 +++
 
 ## Creating visualizations with `altair`
-#### *Build the visualization iteratively*
+#### Build the visualization iteratively
```
joelostblom: No sure why this used to be in italics

trevorcampbell: @joelostblom the other sub-sub-sub-sub section headings in this chapter are also italicized, so this would introduce an inconsistency.

But I think there's a bigger problem -- the translation from R to Python broke things here anyway.

The point of these subheadings wasn't to introduce a new numerical section, it was just to have a "punchline" for each section. 

- One option is to remove this from both versions. 
- (I prefer this one) The other option is to keep it in both but make sure it renders properly in the Py version -- not as a subsection

R version:

image

Py version:

image

trevorcampbell (changes requested):
Need to handle the sub^4 heading problem.

I polished the writing, and it looks like this is the best place to put the note box, so aside from the subheading problem it should be good.

source/viz.md
joelostblom: My bad, I missed that somehow. I can fix the subheading numbering in another PR; it's straightforward from what I recall

trevorcampbell (approved):

PR #245, opened 2023-09-13
Shallower subheading labelling

joelostblom:
as per https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/242

image


trevorcampbell:
Clever! LGTM. I don't see #### in any other doc except reading in the web section, which is also unnumbered in R.

PR #246, opened 2023-09-13
Revert "Shallower subheading labelling"

joelostblom:
@trevorcampbell I just realized something after you merged https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/245. The heading level to number was already set to 3, so we shouldn't need to set it to 2, since the heading we are trying to control is a fourth level heading. I think what actually is happening is that jupyter book does not allow non-consecutive jumps in headings levels. So since we went from a second level to a fourth level heading in the viz chapter, the latter was promoted to a third level heading which used to be numbered in the old config. I don't know if there is a way around this in jupyter book and allow for non-consecutive jumps, but we should revert that PR since it causes an issue with all other third level headings which are not longer numbered
image

### DIFF
--- a/source/_toc.yml
+++ b/source/_toc.yml
@@ -8,7 +8,7 @@ parts:
     - file: acknowledgements.md
     - file: authors.md
 - caption: Chapters
-  numbered: 2
+  numbered: 3
   chapters:
     - file: intro.md
     - file: reading.md
--- a/source/viz.md
+++ b/source/viz.md
@@ -54,7 +54,7 @@ By the end of the chapter, readers will be able to do the following:
 
 ## Choosing the visualization
 
-#### *Ask a question, and answer it*
+<font size="5">*Ask a question, and answer it*</font>
 
 ```{index} question; visualization
 ```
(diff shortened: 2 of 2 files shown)


trevorcampbell:
Ah! Huh. I noticed the 3 vs 4 problem too, but I figured the "chapter-level" headings were implicitly "0", so 3->2 made sense.

But yeah I see this broke some things. Thanks for double checking that.

trevorcampbell (changes requested):
This looks good, but we should try to fix the problem with the viz section here

Can we just use bold+italics for those subheadings? instead of trying to keep them with `#### blah` ?

Or does `{-}` solve it? https://stackoverflow.com/questions/35077571/how-to-remove-heading-numbers-in-jupyter-during-pdf-conversion

joelostblom:
> Or does {-} solve it

Unfortunately it didn't

This is what bold + italics look like. Not the prettiest, but I'm OK with it. Thoughts?

image

trevorcampbell:
We can use `<font size="5">` to do it -- looks like this. I committed it already -- if you're happy with it we can merge

image

trevorcampbell (approved):
LGTM now (also this just becomes italics in the PDF build, which also looks fine -- better than bold+italics IMO)

joelostblom:
Great solution and thanks for updating the PR; merging.

PR #247, opened 2023-09-15
Quote consistency

### LINKED ISSUE #62: Do a pass on quote consistency
Once the book has stabilized, we should make a decision on what kind of string quotes we want to use (single or double) and make it consistent throughout. I tend to use double (for some reason I've long forgotten...I recall that I used to use single ages ago...), but single requires few key presses. I'm happy either way.

trevorcampbell:
Closes #62


trevorcampbell:
@joelostblom this is an annoying finicky change, but figured I may as well just get it over with now. I will scan over the diffs myself too to make sure I didn't break anything, but I'd appreciate an extra set of eyes to double check.

Something to keep a particular eye out for: string formulas that contain quotes, like

```
'blah here is "a thing" blah blah'
```
or
```
"blah here is 'a thing' blah blah"
```

possible I accidentally did

```
"blah here is "a thing" blah blah"
```

which will break things.

joelostblom (approved):
Thanks for taking care of this. i did a  quick review and did not notice any offending cases.

PR #240, opened 2023-09-12
Getting 3D visualizations working again

trevorcampbell:
- Closes #127 
- Closes #83 

One potential issue: I see
```
/home/jovyan/source/classification1.md:786: WARNING: skipping unknown output mime type: application/vnd.plotly.v1+json [mystnb.unknown_mime_type]
```

@joelostblom if you want you can make similar viz adjustments to these new 3d figs that you made previously

In order to test this, just run the docker container interactively (instead of using `build_html.sh`), install `plotly`, and run `jupyter-book build source`.

NB: we will almost certainly have to have a build flag check for the PDF build and use a stored png version of these instead, but we can cross that bridge when we get to it

PR #248, opened 2023-09-15
PDF hover tooltip fix

### LINKED ISSUE #175: Fix render of interactive images in PDF
From @trevorcampbell 's review of https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/160/files

> Is it possible to render a different image in the PDF version of the book that is just a static screenshot of what the tooltip would look like? If not immediately, please open an issue to sort that out (and anywhere else we add interactive figures where it's important that the interactivity is displayed -- though I think it will only matter here in Ch4)

> stuff like "hover over things with the mouse" do not look great in a printed / PDF book

trevorcampbell:
Closes #175 

But I can't test it yet. @joelostblom #171 appears to be a problem still even after the PR in #218 

Was there a regression somewhere? Or did the new Dockerfile build that just happened for #240 break things?


joelostblom:
Hmm, it seems to be working for me with the latest updates on main and a cleared build directory. Any chance there is something in old in your build directory that interferes with building the new pdf file?

trevorcampbell:
You're right -- I needed to clean out my `_build` -- turns out if you build the html copy of the book and then the PDF, the build doesn't display altair plots properly in the PDF. Probably related to jbook caching.

This PR should be good to review now

source/viz.md
```diff
@@ -1100,15 +1101,18 @@ can_lang_plot_tooltip = alt.Chart(can_lang).mark_point(filled=True).encode(
 
 ```{code-cell} ipython3
 :tags: ["remove-cell"]
-# Increasing the dimensions makes all the ticks fit in jupyter book (the fit with the default dimensions in jupyterlab)
-glue('can_lang_plot_tooltip', can_lang_plot_tooltip.properties(height=320, width=420), display=False)
+if 'BOOK_BUILD_TYPE' in os.environ and os.environ['BOOK_BUILD_TYPE'] == 'PDF':
+    glue('can_lang_plot_tooltip', Image('img/viz/languages_with_mouse.png'), display=False)
+else:
+    # Increasing the dimensions makes all the ticks fit in jupyter book (the fit with the default dimensions in jupyterlab)
+    glue('can_lang_plot_tooltip', can_lang_plot_tooltip.properties(height=320, width=420), display=False)
```
joelostblom, line 1108: Nice fix

joelostblom (approved):
lgtm!

trevorcampbell:
@joelostblom in the PDF version the image is too big. I can't seem to shrink it...shrinking the actual file itself didn't help, nor did setting `width=...` in the `Image(...)` and nor did setting `:figwidth:` in the glue directive. I'm a bit stumped...

joelostblom:
Are you able to add a `ppi` attribute to the figure? That is what worked for us in altair with the latex PDFs, see an example of the output here https://github.com/vega/vl-convert/issues/85#issuecomment-1638921581 and here https://github.com/altair-viz/altair/pull/3163. 

If that is not possible, is there any latex directive that could be used? Like setting the image size globally just before this figure and then setting it back?

trevorcampbell:
I've tried a few other things, and nothing seems to work -- either the image disappears entirely, or it appears but is too big.

I am going to merge this anyway because the HTML build is fine, and the PDF build is improved but not perfect yet. Once the changes are limited to just "make the PDF build work for print" I'll return to this.

PR #253, opened 2023-09-21
Remove todo comment

joelostblom:
Looks like this was fixed and we just forgot to delete the comment. From the live book:

image


trevorcampbell:
@joelostblom was the intention of that comment to point out the missing `.columns = [... ]` block at the end?

joelostblom:
I don't know. My assumption was that the intention was to update the pivot code from either a previous version or even the R placeholder. In either case, I don't think adding the `.columns` block to the figure would be that helpful.

PR #262, opened 2023-09-22
Convert static plots to altair

### LINKED ISSUE #147: Change static ggplot figures to altair (e.g. clustering chapter)
There are some pre-rendered ggplot images throughout the book (clustering, for example) that we can replace with altair

### LINKED ISSUE #180: Images of saved charts are from ggplot
At the end of the viz chapter, we compare svg and png by showing an image of both zoomed in. However that image is from the ggplot chart and does not look the same as the python charts we created, so it might be unclear for students why the saved chart changed from blue circles with a white background to black dots with a grey background.

Image in book:
image

Actual saved chart:
image

### LINKED ISSUE #258: Wrong number of cluster points in text
The cluster has 4 points, but:

image

trevorcampbell:
- Closes #147 
- Closes #180 
- Closes #258 
- adds a missing acknowledgment for GCF that we are adding to the R version in an upcoming PR
- removes a remaining R resource from the additional resources section in vsn ctl


joelostblom:
I will go through and assess how much work it is and then I can let you know. When would we need it done by?

trevorcampbell:
It's not blocking right now, but it may start blocking -- let's say within a few weeks?

source/regression2.md
```diff
@@ -305,13 +305,13 @@ small_sacramento_pred = small_sacramento_pred[["sqft", "price", "predicted"]].me
...
-error_plot = small_plot.copy()
-
+v_lines = []
 for i in range(len(small_sacramento)):
     sqft_val = small_sacramento.iloc[i]["sqft"]
     line_df = small_sacramento_pred.query("sqft == @sqft_val")
-    error_plot += alt.Chart(line_df).mark_line(color="red").encode(x="sqft", y="value")
+    v_lines.append(alt.Chart(line_df).mark_line(color="red").encode(x="sqft", y="value"))
 
+error_plot = alt.layer(*v_lines, small_plot).configure_circle(opacity=1)
```
joelostblom, line 314: All the changes in the regression plots is to put the dots on top of the lines which I think looks more appealing

Current:
image

Suggested:
image

source/clustering.md
```diff
@@ -291,7 +291,7 @@ have.
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
-clus = data[data["cluster"] == 2][["bill_length_standardized", "flipper_length_standardized"]]
+clus = data[data["cluster"] == 0][["bill_length_standardized", "flipper_length_standardized"]]
```
joelostblom, line 294: This is the reason for https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/258

source/clustering.md
```diff
@@ -356,14 +394,46 @@ we would compute the WSSD $S^2$ via
 
 $S^2 = \left((x_1 - \mu_x)^2 + (y_1 - \mu_y)^2\right) + \left((x_2 - \mu_x)^2 + (y_2 - \mu_y)^2\right) + \left((x_3 - \mu_x)^2 + (y_3 - \mu_y)^2\right)  +  \left((x_4 - \mu_x)^2 + (y_4 - \mu_y)^2\right)$
 
-These distances are denoted by lines in {numref}`toy-example-clus1-dists-1` for the first cluster of the penguin data example.
+These distances are denoted by lines in {numref}`toy-example-clus1-dists` for the first cluster of the penguin data example.
+
+:::{glue:figure} toy-example-clus1-dists
+:figwidth: 700px
```
joelostblom, line 400: I am not sure this 700px has an effect but kept it in for consistency

source/clustering.md
```diff
@@ -374,37 +444,121 @@ we sum them together to get the *total WSSD*.
...
+            .scale(domain=(-2, 2))
+            .title("Flipper Length (standardized)"),
+        alt.Y("mean(bill_centroid)")
+            .scale(domain=(-2, 2))
+            .title("Flipper Length (standardized)")
+    )
+
+    return (points + centroids).facet(
+        row=alt.Row('iteration', header=alt.Header(title='', labelFontSize=18)),
+        column=alt.Column('update_type', header=alt.Header(title='', labelFontSize=18))
```
joelostblom, line 548: I changed the default layout here to be only in two columns, one for each condition. I think this is much clearer than alternating between the two conditions in four columns, so ideally the same change would be made in the R book, but I wont be able to make it in the near future myself.

I also reduced the number of iterations with 1 since it is already clear what is happening at that point

source/clustering.md
```diff
@@ -374,37 +444,121 @@ we sum them together to get the *total WSSD*.
...
+```
+
+```{code-cell} ipython3
+:tags: [remove-cell]
+
+centroid_init = penguin_data.groupby('label').mean()
+# For some reason, this adjustment is needed to get roughly the same assignment of the points
+# to the clusters as in the R version
+centroid_init.loc[2] = [0.29, 0.36]
+centroid_init.loc[1] = [-0.01, -0.09]
```
joelostblom: I am not sure why this is the case, but I didn't look too closely at the R code (although it seemed like they used the mean there). Not that there is still one point that is different but it has not impact on the take home message

joelostblom (commented):
@trevorcampbell  I made time for this now, since my next window probably won't be until late Dec and it sounds like that would be too late. I think this takes care of point 1. For point 2, I think it would be possible to create that figure in Altair, but I vote for your suggestion to take the path of least resistance and use what we have.

source/regression2.md
trevorcampbell: agreed! nice change

trevorcampbell:
@joelostblom thanks for working on this!

One issue is that the center update / label update figs seem wrong:

image

The orange in the top right definitely isn't getting the right assignments

trevorcampbell:
same issue in the figure right after -- at least the first center/label updates seem off

image

source/clustering.md
trevorcampbell: We did 2 rows here because the single-column format did not look good in the print book

(if you read the text right before the image, it's actually wrong in the R book -- it's the old description where we originally had something more like what you have now -- I've now filed an issue in the R book to fix that text.)

BUT I agree that we can remove the last iteration, in which case maybe it would look fine (and in that case the 2 column format is better).

Let's keep what you have here and I'll change the R version.

trevorcampbell:
@joelostblom I've fixed all the static images now (let's hope I *never* have to do that again. ugh) 

Only thing left to fix on this PR is the bug in the nice new clustering viz you coded up -- any chance you can quickly sort that out?

joelostblom (approved):
@trevorcampbell  Thanks for noticing the issue with the cluster assignment. I wanted you to have a look at that but forgot to comment yesterday. I was perplexed about what was going on there and spent a lot of time trying to figure out without result. It seemed like the cluster centers and labels from KMeans didn't align for some reason. Today I found this in the documentation:

> If the algorithm stops before fully converging (because of tol or max_iter), labels_ and cluster_centers_ will not be consistent, i.e. the cluster_centers_ will not be the means of the points in each cluster. Also, the estimator will reassign labels_ after the last iteration to make labels_ consistent with predict on the training set.

That threw me off, maybe I'm missing something fundamental but I don't understand why all the attributes from the clustering class wouldn't treat the input parameters the same. I tried fixing this by implementing the search manually instead of using KMeans; please have a look and let me know if it checks out.

source/clustering.md
```diff
@@ -374,37 +444,117 @@ we sum them together to get the *total WSSD*.
...
+    2, 2, 1, 1, 0, 0, 0, 1,
+    2, 2, 1, 2, 1, 2,
+    0, 1, 2, 2
+]
+points_kmeans_init = alt.Chart(penguin_data).mark_point(size=75, filled=True, opacity=1).encode(
+    alt.X("flipper_length_standardized").title("Flipper Length (standardized)"),
+    alt.Y("bill_length_standardized").title("Bill Length (standardized)"),
+    alt.Color('label:N').legend(None),
+    alt.Shape('label:N').legend(None).scale(range=['square', 'circle', 'triangle']),
+    alt.Size('label:O').legend(None).scale(type='ordinal', range=[50, 50, 100]),
```
joelostblom, line 478: The reason for the different sizes is that triangle marks seem to be smaller by default and hard to see

trevorcampbell:
looks great, thanks @joelostblom :D merging this now

PR #285, opened 2023-11-11
Caveats of bar plots & plot titles & captions

### LINKED ISSUE #178: Mention when not to use bar charts to avoid common pitfalls
From https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/160#discussion_r1255061215

> **Note:** Although bar charts are often used to display mean and median values,
> they are generally a poor choice for these types of visualizations and should be avoided.
> The reason for this is that they hide all the variation in the data and only show a single value.
> Instead it's better to show the distribution of all the individual data points
> (potentially together with an additional visual mark for the mean or median).

More info here on why this is important and can be misleading https://pages.github.ubc.ca/mds-2022-23/DSCI_531_viz-1_students/lectures/2-data-types_graphical-marks_visual-encondings.html#bar-charts

### LINKED ISSUE #179: Explain how to craft good plot titles and why they are important (R & Py)
From https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/160#discussion_r1255070153. I think titles is an important part of storytelling and should be explained to students. Some more details here https://clauswilke.com/dataviz/figure-titles-captions.html

````
To add some finishing touches to the chart
we will add a title to the chart by specifying `title` argument inside `alt.Chart`.
A good plot title should usually contain the take home message that you want readers of the chart to focus on,
e.g. "The Earth's seven largest landmasses are all continents",
but it could sometimes be more general, e.g. "The twelve largest landmasses on Earth".
Note that plot titles are not always required; e.g. if plots appear as part
of other media (e.g., in a slide presentation, on a poster, in a paper) where
the title may be redundant with the surrounding context.
For categorical encodings,
such as the color and y channels in our chart,
it is often not necessary to include the axis title
as the labels of the categories are enough by themselves.
Particularly in this case where the title clearly states
that we are landmasses,
the titles are redundant and we can remove them.

```{code-cell} ipython3
islands_plot_titled = alt.Chart(
    islands_top12,
    title="The Earth's seven largest landmasses are all continents"
).mark_bar().encode(
    x=alt.X("size",title="Size (1000 square mi)"),
    y=alt.Y("landmass", title="", sort="x"),
    color=alt.Color("landmass_type", title="")
)
```

```{code-cell} ipython3
:tags: ["remove-cell"]
glue('islands_plot_titled', islands_plot_titled, display=True)
```

:::{glue:figure} islands_plot_titled
:figwidth: 700px
:name: islands_plot_titled

Bar plot of size for Earth's largest 12 landmasses with a title.
:::
````

### LINKED ISSUE #287: incorrect caption in viz
Need to go through all the captions to make sure they still make sense, since Altair defaults are different from ggplot defaults and the flow of plot creation changes. For example, in R, we have squished labels in the landmasses example, but in Py we just have a very wide plot, so the caption is wrong

image

trevorcampbell:
- Closes #178 
- Closes #179 
- Closes #287

### DIFF
--- a/source/viz.md
+++ b/source/viz.md
@@ -1271,13 +1271,13 @@ and add a title to the chart by specifying the `title` argument of `alt.Chart`.
 Plot titles are not always required, especially when it would be redundant with an already-existing
 caption or surrounding context (e.g., in a slide presentation with annotations).
 But if you decide to include one, a good plot title should provide the take home message
-that you want readers to focus on, e.g., "Earth's seven largest landmasses are all continents,"
+that you want readers to focus on, e.g., "Earth's seven largest landmasses are continents,"
 or a more general summary of the information displayed, e.g., "The twelve largest landmasses on Earth."
 
 ```{code-cell} ipython3
 islands_plot_sorted = alt.Chart(
 	islands_top12,
-	title="Earth's seven largest landmasses are all continents"
+	title="Earth's seven largest landmasses are continents"
 ).mark_bar().encode(
     x=alt.X("size").title("Size (1000 square mi)"),
     y=alt.Y("landmass").sort("x").title("Landmass"),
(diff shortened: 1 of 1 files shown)


joelostblom:
Yay, great! Thanks for including this!

PR #301, opened 2023-11-12
Handling assign, apply, and lambdas

### LINKED ISSUE #100: Consider removing `apply` and `lambda` completely
Copied from slack:

> What do you think of not teaching apply and lambda at all? It might sound radical at first, but I don't think we need to use these anywhere in dsci 100. The examples we have there right now are better done in other ways, and after spending the last hour trying to come up with better examples, I think all use cases where apply and lambda are justified are too advanced for dsci 100 (and honestly maybe also for MDS based on how I see students often use it). This would cut down the material and reduce complexity at seemingly no/little cost from what I can tell. I can make a PR of what this would look like in all the chapters (we rarely use them already so there would not be that many changes), unless you think it is too preposterous of a suggestion =)

Trevor mentioned this is not the most important thing to do right now so I will hold off with it.

---

Edit: Also note that the way we suggest to use apply makes the syntax more complicated. E.g

```
region_lang.loc[:, "mother_tongue":"lang_known"].apply(lambda col: col.astype("int32"))
```

instead of just

```
region_lang.loc[:, "mother_tongue":"lang_known"].astype("int32")
```

and `apply(max, axis=1` instead of `.max(axis='index')`

### LINKED ISSUE #130: lambda functions in chaining and pandas column attribute shorthand
If I have a df with one column `a`, and I want to do:
- create a new column `b = a+1`
- create a new column `c = b+1`

Then we currently can't chain it. We would have to store an intermediate in order to reference the new column.
```
df = df.assign(b = df['a']+1)
df = df.assign(c = df['b']+1)
```

We can instead create a lambda that gives us a temporary name for the new assigned df https://pandas.pydata.org/docs/user_guide/dsintro.html#assigning-new-columns-in-method-chains

```
df.assign(
        b = df['a'] + 1
   ).assign(
        c = lambda x : x['b'] + 1
)
```

Similarly, we can do `x.b` instead of `x['b']` for brevity. Might be nice to teach and use that where possible.

### LINKED ISSUE #185: Remove .assign in favor of regular column assignment
Most of the time we use `assign` we use it like this in the book:

```
sacramento_train = sacramento_train.assign(column=....)
```

I think that overwriting the same variable name like this is a bit of an antipattern (and likely a spillover from how `mutate` is used in the tidyverse) and that it would be clearer to stick with this type of column assignment instead:

```
sacramento_train[column] = ...
```

This is also the most common column assignment syntax students will see online.

I think this replacement would be trivial in most places in the book, and I can look more in detail if there are some where it wouldn't work if we think a switch could be helpful.

trevorcampbell:
- Closes #185  -- removed `assign` everywhere except in inference, where it is needed inside the list comprehension. I kept the introduction to `assign` in chapter 3, but shortened it and de-emphasized it compared with column assignment.
- Closes #100  -- lambdas and apply are now removed, as they didn't appear anywhere else in the book. I did still keep the section on applying transformations / row-wise operations, because I think it's still very useful material to see. I just used `pandas`'s pre-baked methods, with a `note` box for the `apply` method.
- Closes #130 no longer needed because assign is gone

~~possibly add mutation to ch1~~ turns out this was already done previously (...in fact I think it was me that did it...)

### DIFF
--- a/source/classification1.md
+++ b/source/classification1.md
@@ -628,15 +628,13 @@ Scatter plot of concavity versus perimeter with new observation represented as a
 ```{code-cell} ipython3
 new_obs_Perimeter = 0
 new_obs_Concavity = 3.5
-(
-    cancer
-   [["Perimeter", "Concavity", "Class"]]
-   .assign(dist_from_new = (
+cancer["dist_from_new"] = (
        (cancer["Perimeter"] - new_obs_Perimeter) ** 2
      + (cancer["Concavity"] - new_obs_Concavity) ** 2
-   )**(1/2))
-   .nsmallest(5, "dist_from_new")
-)
+)**(1/2)
+cancer.nsmallest(5, "dist_from_new")[
+	["Perimeter", "Concavity", "Class", "dist_from_new"]
+]
 ```
 
 ```{code-cell} ipython3
@@ -751,16 +749,14 @@ three predictors.
 new_obs_Perimeter = 0
 new_obs_Concavity = 3.5
 new_obs_Symmetry = 1
-(
-    cancer
-    [["Perimeter", "Concavity", "Symmetry", "Class"]]
-    .assign(dist_from_new = (
-        (cancer["Perimeter"] - new_obs_Perimeter) ** 2
-        + (cancer["Concavity"] - new_obs_Concavity) ** 2
-        + (cancer["Symmetry"] - new_obs_Symmetry) ** 2
-    )**(1/2))
-    .nsmallest(5, "dist_from_new")
-)
+cancer["dist_from_new"] = (
+      (cancer["Perimeter"] - new_obs_Perimeter) ** 2
+    + (cancer["Concavity"] - new_obs_Concavity) ** 2
+    + (cancer["Symmetry"] - new_obs_Symmetry) ** 2
+)**(1/2)
+cancer.nsmallest(5, "dist_from_new")[
+    ["Perimeter", "Concavity", "Symmetry", "Class", "dist_from_new"]
+]
 ```
 
 Based on $K=5$ nearest neighbors with these three predictors we would classify 
--- a/source/classification1.md
+++ b/source/classification1.md
@@ -632,9 +632,12 @@ cancer["dist_from_new"] = (
        (cancer["Perimeter"] - new_obs_Perimeter) ** 2
      + (cancer["Concavity"] - new_obs_Concavity) ** 2
 )**(1/2)
-cancer.nsmallest(5, "dist_from_new")[
-	["Perimeter", "Concavity", "Class", "dist_from_new"]
-]
+cancer.nsmallest(5, "dist_from_new")[[
+	"Perimeter", 
+	"Concavity", 
+	"Class", 
+	"dist_from_new"
+]]
 ```
 
 ```{code-cell} ipython3
@@ -754,9 +757,13 @@ cancer["dist_from_new"] = (
     + (cancer["Concavity"] - new_obs_Concavity) ** 2
     + (cancer["Symmetry"] - new_obs_Symmetry) ** 2
 )**(1/2)
-cancer.nsmallest(5, "dist_from_new")[
-    ["Perimeter", "Concavity", "Symmetry", "Class", "dist_from_new"]
-]
+cancer.nsmallest(5, "dist_from_new")[[
+    "Perimeter", 
+    "Concavity", 
+    "Symmetry", 
+    "Class", 
+    "dist_from_new"
+]]
 ```
 
 Based on $K=5$ nearest neighbors with these three predictors we would classify 
--- a/source/classification1.md
+++ b/source/classification1.md
@@ -633,10 +633,10 @@ cancer["dist_from_new"] = (
      + (cancer["Concavity"] - new_obs_Concavity) ** 2
 )**(1/2)
 cancer.nsmallest(5, "dist_from_new")[[
-	"Perimeter", 
-	"Concavity", 
-	"Class", 
-	"dist_from_new"
+    "Perimeter", 
+    "Concavity", 
+    "Class", 
+    "dist_from_new"
 ]]
 ```
 
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -1025,12 +1025,12 @@ sacr_results = pd.DataFrame(sacr_gridsearch.cv_results_)
 sacr_results["sem_test_score"] = sacr_results["std_test_score"] / 5**(1/2)
 sacr_results["mean_test_score"] = -sacr_results["mean_test_score"]
 sacr_results = (
-	sacr_results[[
-		"param_kneighborsregressor__n_neighbors",
-		"mean_test_score",
-		"sem_test_score"
-	]]
-	.rename(columns={"param_kneighborsregressor__n_neighbors" : "n_neighbors"})
+    sacr_results[[
+        "param_kneighborsregressor__n_neighbors",
+        "mean_test_score",
+        "sem_test_score"
+    ]]
+    .rename(columns={"param_kneighborsregressor__n_neighbors" : "n_neighbors"})
 )
 
 # show only the row of minimum RMSPE
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -899,10 +899,10 @@ base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
 
 # Add the predictions as a line
 sacr_preds_plot = base_plot + alt.Chart(
-	sqft_prediction_grid, 
-	title=f"K = {best_k_sacr}"
+    sqft_prediction_grid, 
+    title=f"K = {best_k_sacr}"
 ).mark_line(
-	color="#ff7f0e"
+    color="#ff7f0e"
 ).encode(
     x="sqft",
     y="predicted"
--- a/source/regression1.md
+++ b/source/regression1.md
@@ -294,16 +294,14 @@ of a house that is 2,000 square feet.
 ```
 
 ```{code-cell} ipython3
-nearest_neighbors = (
-    small_sacramento.assign(diff=(2000 - small_sacramento["sqft"]).abs())
-    .nsmallest(5, "diff")
-)
-
-nearest_neighbors
+small_sacramento["dist"] = (2000 - small_sacramento["sqft"]).abs()
+small_sacramento.nsmallest(5, "dist")
 ```
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
+
+nearest_neighbors = small_sacramento.nsmallest(5, "dist")
 
 nn_plot = small_plot + rule
 
@@ -609,16 +607,15 @@ sacr_gridsearch.fit(
 )
 
 # Retrieve the CV scores
-sacr_results = pd.DataFrame(sacr_gridsearch.cv_results_)[[
-    "param_kneighborsregressor__n_neighbors",
-    "mean_test_score",
-    "std_test_score"
-]]
+sacr_results = pd.DataFrame(sacr_gridsearch.cv_results_)
+sacr_results["sem_test_score"] = sacr_results["std_test_score"] / 5**(1/2)
 sacr_results = (
-    sacr_results
-    .assign(sem_test_score=sacr_results["std_test_score"] / 5**(1/2))
+    sacr_results[[
+        "param_kneighborsregressor__n_neighbors",
+        "mean_test_score",
+        "sem_test_score"
+    ]]
     .rename(columns={"param_kneighborsregressor__n_neighbors": "n_neighbors"})
-    .drop(columns=["std_test_score"])
 )
 sacr_results
 ```
@@ -834,12 +831,10 @@ model uses a different default scoring metric than the RMSPE.
 ```{code-cell} ipython3
 from sklearn.metrics import mean_squared_error
 
-sacr_preds = sacramento_test.assign(
-    predicted = sacr_gridsearch.predict(sacramento_test)
-)
+sacramento_test["predicted"] = sacr_gridsearch.predict(sacramento_test)
 RMSPE = mean_squared_error(
-    y_true = sacr_preds["price"],
-    y_pred=sacr_preds["predicted"]
+    y_true = sacramento_test["price"],
+    y_pred = sacramento_test["predicted"]
 )**(1/2)
 RMSPE
 ```
@@ -890,9 +885,7 @@ sqft_prediction_grid = pd.DataFrame({
     "sqft": np.arange(sacramento["sqft"].min(), sacramento["sqft"].max(), 10)
 })
 # Predict the price for each of the sqft values in the grid
-sacr_preds = sqft_prediction_grid.assign(
-    predicted = sacr_gridsearch.predict(sqft_prediction_grid)
-)
+sqft_prediction_grid["predicted"] = sacr_gridsearch.predict(sqft_prediction_grid)
 
 # Plot all the houses
 base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
@@ -905,11 +898,14 @@ base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
 )
 
 # Add the predictions as a line
-sacr_preds_plot = base_plot + alt.Chart(sacr_preds, title=f"K = {best_k_sacr}").mark_line(
-    color="#ff7f0e"
+sacr_preds_plot = base_plot + alt.Chart(
+	sqft_prediction_grid, 
+	title=f"K = {best_k_sacr}"
+).mark_line(
+	color="#ff7f0e"
 ).encode(
-    x="sqft",
-    y="predicted"
+	x="sqft",
+	y="predicted"
 )
 
 sacr_preds_plot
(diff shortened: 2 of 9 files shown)


source/classification1.md
```diff
@@ -628,15 +628,16 @@ Scatter plot of concavity versus perimeter with new observation represented as a
...
      + (cancer["Concavity"] - new_obs_Concavity) ** 2
-   )**(1/2))
-   .nsmallest(5, "dist_from_new")
-)
+)**(1/2)
+cancer.nsmallest(5, "dist_from_new")[[
+	"Perimeter", 
+	"Concavity", 
+	"Class", 
+	"dist_from_new"
```
joelostblom:
```suggestion
    "Perimeter", 
    "Concavity", 
    "Class", 
    "dist_from_new"
```

source/wrangling.md
```diff
@@ -800,7 +800,10 @@ It makes sense for `region`, `category`, and `language` to be stored as an
 above a numeric threshold of a column).
 That won't be possible if the variable is stored as a `object`.
```
joelostblom:
```suggestion
That won't be possible if the variable is stored as an `object`.
```

source/wrangling.md
```diff
@@ -800,7 +800,10 @@ It makes sense for `region`, `category`, and `language` to be stored as an
 above a numeric threshold of a column).
```
joelostblom, line 799: Minor: On line 785 above we refer to the object dtype as "object" with quotation marks, but further than we use backticks: `object` so maybe change the above to that too for consistency?

joelostblom: I would also change:

> It makes sense for `region`, `category`, and `language` to be stored as an
`object` type.

to

> It makes sense for `region`, `category`, and `language` to be stored as an
`object` type since they hold categorical values.

source/wrangling.md
```diff
@@ -1449,44 +1452,23 @@ region_lang.value_counts("region", normalize=True)
...
+A transformation applied across many columns. The darker, top row of each table represents the column headers.
 ```
 
-+++
-
-For example,
-imagine that we wanted to convert all the numeric columns
+For example, imagine that we wanted to convert all the numeric columns
 in the `region_lang` data frame from `int64` type to `int32` type
 using the `.as_type` function.
```
joelostblom:
```suggestion
using the `.astype` function.
```

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
-```
-
-```{note}
-Because we didn't give the `lambda` function a name, we have to surround it with
-parentheses too if we want to call it. Otherwise, if we wrote something like `lambda x: 2*x(2)`, Python would get confused
-and think that `(2)` was part of the instructions that comprise the `lambda` function.
-As long as we don't want to call the `lambda` function ourselves, we don't need those parentheses. For example,
-we can pass a `lambda` function as an argument to `apply` without any parentheses.
-```
+We can simply call the `.as_type` function to apply it across the desired range of columns.
```
joelostblom:
```suggestion
We can simply call the `.astype` function to apply it across the desired range of columns.
```

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
+`most_at_home`, `most_at_work` and `lang_known` for each row in the data frame. It
+is often the case that we want to include a column result
+from a row-wise operation as a new column in the data frame, so that we can make
 plots or continue our analysis. To make this happen,
-we will use `assign` to create a new column. This is discussed in the next section.
+we will use column assignment or the `assign` function to create a new column. 
+This is discussed in the next section.
+
+```{note}
+While `pandas` provides many methods (like `max`, `as_type`, etc.) that can be applied to a data frame,
```
joelostblom:
```suggestion
While `pandas` provides many methods (like `max`, `astype`, etc.) that can be applied to a data frame,
```

source/wrangling.md
```diff
@@ -800,7 +800,10 @@ It makes sense for `region`, `category`, and `language` to be stored as an
 above a numeric threshold of a column).
 That won't be possible if the variable is stored as a `object`.
 Fortunately, the `pandas.to_numeric` function provides a natural way to fix problems
```
joelostblom: In section 3.10, we show them `astype`, should we use that here also for consistency? Then we could also convert both columns together in a single line instead of two below.

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
 In other words, we want to apply the `max` function *row-wise.*
-In order to tell `apply` that we want to work row-wise (as opposed to acting on each column
+In order to tell `max` that we want to work row-wise (as opposed to acting on each column
 individually, which is the default behavior), we just specify the argument `axis=1`.
-For example, in the case of the `max` function, this tells Python that we would like
-the `max` within each row of the input, as opposed to being applied on each column.
 
 ```{code-cell} ipython3
-region_lang_nums.apply(max, axis=1)
+region_lang_nums.max(axis=1)
```
joelostblom, line 1515: An alternative here is to say `axis='columns'` and then emphasize above that when we are applying a function row-wise it will go across all the *columns* which links to the name of this option (so that students don't think it should be called "rows" since it is "row-wise"). I'll let you pick what you think is clearer, just a possible alternative, I generally like naming the axes with a string instead of a number personally since I forget which is axis 0 and 1 constantly... (and I think there is some pandas function where they are confusingly inverted). There is a couple of occurences of this further down in case you decide to change it

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
-region_lang_nums = region_lang.loc[:, "mother_tongue":"lang_known"].apply(lambda col: col.astype("int32"))
+region_lang_nums = region_lang.loc[:, "mother_tongue":"lang_known"].astype("int32")
 region_lang_nums.info()
 ```
-You can now see that the columns from `mother_tongue` to `lang_known` are type `int32`.
-You can also see that `apply` returns a data frame with the same number of columns and rows
-as the input data frame. The only thing `apply` does is use the `lambda` function argument
-on each of the specified columns.
-
-### Apply a function row-wise with `apply`
```
joelostblom, line 1549: I think this section is much clearer now, nice!

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
 
 
 ```{index} pandas.DataFrame; []
 ```
 
-### Using `assign` to create new columns
-
-When we compute summary statistics with `agg` or apply functions using `apply`
+When we compute summary statistics with `agg` or apply functions using `apply`,
 those give us new data frames. But what if we want to append that information
```
joelostblom:
```suggestion
When we compute summary statistics or apply functions,
a new data frame or series is created. But what if we want to append that information
```

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
+and to append that as an additional column of the `region_lang` data frame.
+In this case, we have two options: we can either  create a new column within the `region_lang` data
+frame itself, or create an entirely new data frame
+with the `assign` method. The first option we have seen already in earlier chapters, and is
+the more commonly used pattern in practice:
+```{code-cell} ipython3
+:tags: ["output_scroll"]
+region_lang["maximum"] = region_lang_nums.apply(max, axis=1)
+region_lang
+```
```
joelostblom:
````suggestion
to an existing data frame? For example, say we wanted to compute the
maximum value in each row of the `region_lang_nums` data frame,
and to append that as an additional column of the `region_lang` data frame.
In this case, we have two options: we can either  create a new column within the `region_lang` data
frame itself, or create an entirely new data frame
with the `assign` method. The first option we have seen already in earlier chapters, and is
the more commonly used pattern in practice:
```{code-cell} ipython3
:tags: ["output_scroll"]
region_lang["maximum"] = region_lang_nums.max(axis=1)
region_lang
```
````

No reason to use `apply`  here either

File: source/wrangling.md [hunk 12]
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
+You can see above that the `region_lang` data frame now has an additional column named `maximum`.
+The `maximum` column contains
+the maximum value between `mother_tongue`,
+`most_at_home`, `most_at_work` and `lang_known` for each language
+and region, just as we specified! 
+
+To instead create an entirely new data frame with the `assign` method, we specify one argument for each column we want to create.
```
joelostblom:
```suggestion
To instead create an entirely new data frame, we can use the `assign` method and specify one argument for each column we want to create.
```

source/wrangling.md
```diff
@@ -1500,109 +1482,81 @@ region_lang
...
 In this case we want to create one new column named `maximum`, so the argument
 to `assign` begins with `maximum = `.
```
joelostblom:
```suggestion
to `assign` begins with `maximum= `.
```

source/wrangling.md
```diff
@@ -1614,17 +1568,17 @@ region_lang.assign(
   maximum = region_lang_nums.apply(max, axis=1)
```
joelostblom:
```suggestion
  maximum=region_lang_nums.max(axis=1)
```

source/wrangling.md
```diff
@@ -1614,17 +1568,17 @@ region_lang.assign(
...
 )
 ```
-This gives us a new data frame that looks like the `region_lang` data frame,
-except that it has an additional column named `maximum`.
-The `maximum` column contains
-the maximum value between `mother_tongue`,
-`most_at_home`, `most_at_work` and `lang_known` for each language
-and region, just as we specified!
+This data frame looks just like the previous one, except that it is a copy of `region_lang`, not `region_lang` itself; making further
+changes to this data frame will not impact the original `region_lang` data frame.
```
joelostblom, line 1572: To make this even more explicit, should we assign it to a new variable name above?

source/wrangling.md
```diff
@@ -1676,21 +1630,19 @@ five_cities
...
 ```{code-cell} ipython3
 :tags: ["output_scroll"]
-english_lang = english_lang.assign(
+english_lang.assign(
   city_pops=[4098927, 5928040, 1392609, 1321426, 2463431]
 )
-english_lang
 ```
 
 ```{note}
```
joelostblom: It's good that we have this note, I was just thinking about it when reading through the passage above. Can we change:

> We do it here to demonstrate another usage of `assign` that does not involve `apply`.

to

> We do it here to demonstrate another usage of `assign` and regular column assignment.

I also think we can move it after do the regular column assignment and point out that it is an issue with both assign and regular column assignment. Then the note will also act as a break before the proportion section.

source/wrangling.md
```diff
@@ -1676,21 +1630,19 @@ five_cities
 ```
 The data frame above shows that the populations of the five cities in 2016 were
 5928040 (Toronto), 4098927 (Montréal),  2463431 (Vancouver), 1392609 (Calgary), and 1321426 (Edmonton).
-We will add this information to our data frame in a new column named `city_pops` by using `assign`.
+We will first add this information to a new data frame in a new column named `city_pops` by using `assign`.
```
joelostblom:
```suggestion
Next, we will add this information to a new data frame column called `city_pops`.
Once again, we will illustrate how to do this using both regular column assignment
and the  `assign` method, starting with the latter.
```

Just being a bit more explicit in what's happening next

source/wrangling.md
```diff
@@ -1701,96 +1653,78 @@ one would solve this problem in a less error-prone way using
 the `merge` function, which lets you combine two data frames. We will show you an
 example using `merge` at the end of the chapter!
 ```
-
-Now we have a new column with the population for each city. Finally, we calculate the
-proportion of people who speak English the most at home by taking the ratio of the columns
-`most_at_home` and `city_pops`. We will again add this to our data frame using `assign`.
+Once again we can also instead directly modify the `english_lang` data frame using column assignment.
```
joelostblom:
```suggestion
Instead of using the `assign` method we can directly modify the `english_lang` data frame using regular column assignment.
This would be a more natural choice in this particular case,
since the syntax is more convenient for simple column modifications and additions.
```

source/wrangling.md
```diff
@@ -1701,96 +1653,78 @@ one would solve this problem in a less error-prone way using
...
-```{figure} img/wrangling/pandas_assign_args_labels.png
-:name: fig:img-assign
-:figclass: figure
-
-Syntax for the `assign` function.
+Wait a moment...what is that warning message? It seems to suggest that something went wrong, but
+if we inspect the `english_lang` data frame above, it looks like the city populations were added 
+just fine! As it turns out, this is caused by the earlier filtering we did from `region_lang` to
+produce the original `english_lang`. The details are a little bit technical, but
+`pandas` sometimes does not like it when you subset a data frame using `[]` or `loc[]` followed by
```
joelostblom, line 1676: Great to explain this!

source/wrangling.md
```diff
@@ -1701,96 +1653,78 @@ one would solve this problem in a less error-prone way using
...
-
+Now we have a new column with the population for each city. Finally, we calculate the
+proportion of people who speak English the most at home by taking the ratio of the columns
+`most_at_home` and `city_pops`. Let's modify the `most_at_home` column directly; in this case
+we can just assign directly to the column.
+This is precisely what we did in {numref}`str-split`,
+when we first read in the `"region_lang_top5_cities_messy.csv"` data and we needed to convert a few
+of the variables to numeric types.
+Note that it is again possible to instead use the `assign` function to produce a new data frame when modifying an existing column,
+although this is not commonly done.
```
joelostblom: I think it is a bit odd that we end up `most_at_home` as a rate and `most_at_work` as an absolute number, but with column names that imply that they are the same measure in different context. I would either create a new column with `rate` appended or convert both to rates.

source/wrangling.md
```diff
@@ -1701,96 +1653,78 @@ one would solve this problem in a less error-prone way using
...
-the power to overwrite the entries of a column, it is a good idea to create a new data frame object so that if
-you make a mistake, you can start again from the original data frame.
-
 +++
 
 ## Using `merge` to combine data frames
 
+Let's return to the situation right before we added the city populations
+of Toronto, Montréal, Vancouver, Calgary, and Edmonton to the `english_lang` data frame. Before adding the new column, we had filtered
+our `region_lang` to create the `english_lang` data frame containing only English speakers in the five cities
```
joelostblom:
```suggestion
`region_lang` to create the `english_lang` data frame containing only English speakers in the five cities
```

source/wrangling.md
joelostblom, line 1: Overall I think this is less complicated now and easier to understand! Should we somehow include regular column assignment as a learning objective too?

source/regression1.md
```diff
@@ -294,17 +294,15 @@ of a house that is 2,000 square feet.
...
-nearest_neighbors
+small_sacramento["dist"] = (2000 - small_sacramento["sqft"]).abs()
+small_sacramento.nsmallest(5, "dist")
 ```
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
 
+nearest_neighbors = small_sacramento.nsmallest(5, "dist")
+
```
joelostblom:
````suggestion
small_sacramento["dist"] = (2000 - small_sacramento["sqft"]).abs()
nearest_neighbors = small_sacramento.nsmallest(5, "dist")
nearest_neighbors
```

```{code-cell} ipython3
:tags: [remove-cell]


````

We use `nearest_neighbors` in the next visible cell, so we should show how we define it

File: source/regression1.md [hunk 23]
```diff
@@ -905,11 +898,14 @@ base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
 )
 
 # Add the predictions as a line
-sacr_preds_plot = base_plot + alt.Chart(sacr_preds, title=f"K = {best_k_sacr}").mark_line(
-    color="#ff7f0e"
+sacr_preds_plot = base_plot + alt.Chart(
+	sqft_prediction_grid, 
+	title=f"K = {best_k_sacr}"
+).mark_line(
+	color="#ff7f0e"
```
joelostblom:
```suggestion
sacr_preds_plot = base_plot + alt.Chart(
    sqft_prediction_grid, 
    title=f"K = {best_k_sacr}"
).mark_line(
    color="#ff7f0e"
```

source/regression1.md
```diff
@@ -905,11 +898,14 @@ base_plot = alt.Chart(sacramento).mark_circle(opacity=0.4).encode(
...
 ).encode(
-    x="sqft",
-    y="predicted"
+	x="sqft",
+	y="predicted"
```
joelostblom:
```suggestion
    x="sqft",
    y="predicted"
```

source/regression1.md
```diff
@@ -1018,25 +1014,24 @@ sacr_gridsearch = GridSearchCV(
...
-    sacr_results
-    .assign(sem_test_score=sacr_results["std_test_score"] / 5**(1/2))
-    .rename(columns={"param_kneighborsregressor__n_neighbors" : "n_neighbors"})
-    .drop(columns=["std_test_score"])
+	sacr_results[[
+		"param_kneighborsregressor__n_neighbors",
+		"mean_test_score",
+		"sem_test_score"
+	]]
+	.rename(columns={"param_kneighborsregressor__n_neighbors" : "n_neighbors"})
```
joelostblom:
```suggestion
    sacr_results[[
        "param_kneighborsregressor__n_neighbors",
        "mean_test_score",
        "sem_test_score"
    ]]
    .rename(columns={"param_kneighborsregressor__n_neighbors" : "n_neighbors"})
```

joelostblom (commented):
Big :rocket: for this PR! I'm happy to see this change in the book and thinks it will be helpful for students as it is more intuitive now in my opinion. I have some comments, but nothing major (except for some last remaining apply instances), so feel free to take an leave what you want.

source/wrangling.md
trevorcampbell: Yes -- will handle that in https://github.com/UBC-DSCI/introduction-to-datascience/issues/504

trevorcampbell: (that issue is on the R book but it's for both Py and R)

source/wrangling.md
trevorcampbell: will fix both

source/wrangling.md
trevorcampbell: agree and done

source/wrangling.md
trevorcampbell: I'll open this as a low-priority issue -- I'd need to think more carefully about whether/how to propagate this among chapters, and I don't think it's necessary for 1st edition.

source/wrangling.md
trevorcampbell: agree done

source/wrangling.md
trevorcampbell: fine as-is

source/wrangling.md
trevorcampbell: agree fixed (and bonus assignment to loc)

PR #309, opened 2023-11-12
Consistency of usage of `_best_params` and `.fit`

### LINKED ISSUE #134: Inconsistent treatment of `.fit` output
Sometimes we store the output in a `blah_fit` named object, sometimes we just use the fact that `.fit` modifies the object in-place. I think the in-place modification is a bit unusual since pandas does everything not in-place by default, so we should either point it out somewhere or do something smarter.

(this applies to other pipeline outputs -- not just fit)

### LINKED ISSUE #205: Add `best_params_` to classification 2?
We recently edited the Python version to avoid using `best_estimator_` and instead refer to `best_params_` whenever necessary (with `.predict` on the fit object).

Might be a good idea to move this back to classification 2 (in the R version too), but that requires investigation on whether it's possible (we do some manual manipulation of scoring in the Py version) and whether it would work well in the R version too.

_Originally posted by @trevorcampbell in https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/201#discussion_r1298826595_

### LINKED ISSUE #233: Be consistent in how we fit models
The textbook and assignments have a variety of ways to fit models currently, including:

```
m = M()
m.fit()
```

```
m = M()
m = m.fit()
```

```
m = M().fit
```

I think this can be confusing and it would be nice if we could standardize on one main approach. My vote would be on the first one, since it separates the initialization and fitting into two separate steps and avoids reassignment of the fitted object which makes it clear it is modified in place. There are a few instances with list comprehensions where we still need the third approach and we can make this explicit when we use it "here we initialize and train the model in one line". The second approach could be avoided completely.

### LINKED ISSUE #281: Show `best_params_` for classification?
in the Python version, for classification we don't save the object that we would ask for `best_params_` from, so unless students read the regression chapter and extrapolate to that they will have a hard time. They will think they need to make a data frame of all the parameters and their scores, and sort it and grab the value.

In some sense, it is good to get the values from the data frame, as you should really look at more than just one number, but if you do agree with the best param suggested by the CV procedure, then programmatically its nice to have these.

trevorcampbell:
- Closes #233 
- Closes #134 
- Closes #281 
- Closes #205


joelostblom:
:rocket:

PR #314, opened 2023-11-13
Various improvements to predictive chapters

### LINKED ISSUE #135: Different train/test split in Reg1 and Reg2
This causes the RMSE in the KNN vs Linear comparison in reg 2 to not line up properly

Also the split proportion in Reg 1 is 0.75, in Reg 2 is 0.6...

Check this for classification 1 and 2 too, in case it matters.

### LINKED ISSUE #212: Show regression predictions for test data like in the assignments instead of for all data?
Details in https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/207/files#r1299125230

### LINKED ISSUE #231: Talk more about goodness of fit or change worksheet
One of the regression assignments has a question of whether R2 is measure of goodness of fit or prediction error, but we odn't describe the differences in the textbook and barely mention goodness of fit at all

### LINKED ISSUE #280: Need "Evaluating on the test set" section for classification chapters
In the classification chapter, after you do CV to select your parameters, then you need to refit/retrain on the entire training dataset for the chosen parameter(s) before evaluating on the test data set. In the book we don't do this in the classification case! Only the regression case...

I think this is not great - what if someone only reads our classification chapters and misses this? Like those two classification chapters are where we set the stage for how to do this correctly.

Like I think we need this section in the classification chapter:
https://datasciencebook.ca/regression1.html#evaluating-on-the-test-set

I think it should go right above here: https://datasciencebook.ca/regression1.html#evaluating-on-the-test-set

trevorcampbell:
- Closes #280 
- Closes #135

### DIFF
--- a/source/classification2.md
+++ b/source/classification2.md
@@ -675,7 +675,7 @@ from sklearn.metrics import recall_score, precision_score
 precision_score(
     y_true=cancer_test["Class"],
     y_pred=cancer_test["predicted"],
-    pos_label='Malignant'
+    pos_label="Malignant"
 )
 ```
 
@@ -683,7 +683,7 @@ precision_score(
 recall_score(
     y_true=cancer_test["Class"],
     y_pred=cancer_test["predicted"],
-    pos_label='Malignant'
+    pos_label="Malignant"
 )
 ```
 The output shows that the estimated precision and recall of the classifier on the test
--- a/source/classification2.md
+++ b/source/classification2.md
@@ -491,7 +491,7 @@ right proportions of each category of observation.
 
 ```{code-cell} ipython3
 :tags: [remove-cell]
-# seed hacking to get a split so that recall goes up when we tune later on
+# seed hacking
 np.random.seed(1)
 ```
 
--- a/source/classification2.md
+++ b/source/classification2.md
@@ -717,12 +717,12 @@ acc_eq_math = Math(acc_eq_str)
 glue("acc_eq_math_glued", acc_eq_math)
 
 prec_eq_str = r"\mathrm{precision} = \frac{\mathrm{number \; of  \; correct  \; positive \; predictions}}{\mathrm{total \;  number \;  of  \; positive \; predictions}} = \frac{"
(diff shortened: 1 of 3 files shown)


joelostblom:
Nice! I think the added section is clear and will be helpful for students. One thing I would add is a comment on recall and precision compared to the un-tuned classifier, which are actually slightly different (in an undesirable way for recall). I also noticed that we did an error in our computation of recall further up and that we never show how to use sklearn to get these numbers; we only do it manually via the confusion matrix. Details with screenshots:

1. **Mix up of positive label.** In 6.3, we correctly define "Malignant" as our positive label, since that is what we are looking for:
    image

    However, in 6.5.5, we use "Benign" as the possible label when we manually compute precision and recall
    image

2. **Not showing how to compute precision and recall with sklearn**. In 6.5.5 We show how to compute accuracy via `.score`, but we never show how to compute recall and precision, instead we compute these manually from the confusion matrix:
    image

    I think we could consider changing that paragraph to:

    > The output shows that the estimated accuracy of the classifier on the test data was 88%. To compute the precision and recall we can use the following functions from scikit-learn:
    > `code cell with precision/recall computation`
    > We can see that our precision was .... and our recall was ... . Finally, we can also look at the confusion matrix for the classifier using the crosstab function from pandas. The crosstab function takes two arguments: the actual labels first, then the predicted labels second. *The columns and rows are ordered alphabetically, but our positive label is still "Malignant", even if it is not in the top left corner as in the general confusion matrix above*.

3. **Comment on precision and/or recall after tuning**.

    Before tuning:
    image

    After tuning:
    image

    I think we can add a comment on that although accuracy remain similar, our classifier now has slightly worse recall and misses more malignant samples (TPs), and briefly remind students why TPs are important in our context and that we might think more carefully about how to choose our optimal hyperparameters. Optionally we can also show the sklearn functions to compute the scores here in addition to doing it manually.

trevorcampbell:
@joelostblom all great comments (and I'm not sure how I missed that benign vs malignant issue in the equations...probably when we did the transpose of the matrix I forgot to adjust those glues too)

Thanks a lot -- will fix those and merge.

